PR #75, opened 2018-03-30
Grammar pass

MattWis:
Add some words to make the newsletter read more clearly.

Thanks for all of the work in the embedded-wg! It's really exciting to see how far we've come, and all of the wonderful places we'll go!


jamesmunns:
Looks good to me, thanks @MattWis!

PR #81, opened 2018-04-08
Consistency cleanup for the books

jamesmunns:
* Creates a subdirectory, `books/` to house both projects
* Updates the `License` field of the READMEs of both projects to be more consistent
* Relicenses the `embedonomicon` from CC0 to CC-BY-SA 4.0 (requires signoff by @japaric)
* Minor README changes to both projects for consistency

@japaric could you please make a comment to this PR that you accept relicencing from CC0 to CC-BY-SA v4.0, and also review the changes that I made to both READMEs?

PR #82, opened 2018-04-08
embedonomicon: remove mentions of Xargo and the intrinsics / compiler_builtins chapters

japaric:
depends on #81 

r? @jamesmunns


jamesmunns:
Also for `compiler-builtins` and `intrinsics`, would it make sense to leave the chapters in, or maybe rewrite them from the perspective of someone who has/wants to replace the defaults provided now in the cargo component?

Is that a thing you could see people wanting to do? Or is it necessary for targets that don't have a published component, like `risc-v` or `avr`?

whitequark:
Yes, please leave the xargo and compiler-builtins chapters in! Not all embedded development is Cortex-M...

jamesmunns:
@japaric could you please make follow-up issues about the new content you described? I will go ahead and merge.

PR #114, opened 2018-07-06
Fix "2018 Edition" nomenclature

johnthagen:
See: https://blog.rust-lang.org/2018/03/12/roadmap.html


jamesmunns:
The books no longer live in this repo. Closing the PR, Thank you @johnthagen!

PR #127, opened 2018-07-17
Fix link to Rust RFC process in README

johnthagen:
Open question: Is this the correct link that we want?

Another possibility would be the raw GitHub link: https://github.com/rust-lang/rfcs


japaric:
Yes, thanks @johnthagen.

PR #136, opened 2018-07-30
[RFC] rust-embedded teams and maintenance of the embedded ecosystem

japaric:
closes #46 

[Rendered](https://github.com/rust-lang-nursery/embedded-wg/blob/teams/rfcs/0000-teams.md).

RFCs for creating each team

- MSP430 team #137
- RISCV team #138 
- embedded Linux team #139 
- resources team #140 
- tools team #141 
- HAL team #142 
- Cortex-M team #143 
- triage team #144

PR #140, opened 2018-07-30
[RFC] Bootstrap the resources team

japaric:
As per RFC #136 this PR kickstarts the resources team

Note that the team composition is not final. The team is expected to grow over time. The mechanism
for joining a team is specified in RFC #136.

@jamesmunns approve this PR to accept my invitation to this team

@therealprof you have been doing a great job maintaining awesome embedded Rust; would you be
interested in joining this team to help review work on the books?

@andre-richter before, you expressed interest in joining this team; would you be interested in
joining to help review work on the books?

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,7 @@ The resources team develops, maintains and curates resources on embedded Rust.
 
 - [@japaric]
 - [@jamesmunns]
+- [@andre-richter](https://github.com/andre-richter)
 
 #### Projects
 
--- a/README.md
+++ b/README.md
@@ -41,6 +41,24 @@ What is it that we really want? At a broad level:
 [@pftbest]: https://github.com/pftbest
 [@thejpster]: https://github.com/thejpster
 
+### The resources team
+
+The resources team develops, maintains and curates resources on embedded Rust.
+
+#### Members
+
+- [@japaric]
+- [@jamesmunns]
+
+#### Projects
+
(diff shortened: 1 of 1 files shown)


andre-richter:
Yes I would still be very interested in doing reviews.

jamesmunns:
@japaric do you want to include the [twitter account] in with the resources? I don't know where it would better fit, other than directly with `rust-embedded/wg.`

japaric:
@jamesmunns @andre-richter awesome! Thanks.

@jamesmunns having twitter under the team sounds reasonable to me. What do you think, @andre-richter?

andre-richter:
Thanks for adding! 😃

Regarding Twitter, I'd say depends on the scope of what it should cover. If all teams should be free to announce for example new versions of their tools themselves, then including it here instead of the top of might introduce an unessecary bottleneck?

therealprof:
@japaric Yes, please.

japaric:
@therealprof awesome, added.

@andre-richter so far we have been using the twitter account to tweet the newsletters, and have been retweeting help wanted stuff and project announcements. Now that there several projects under the rust-embedded org I think it makes sense to use the org twitter account to tweet announcements of new releases of those projects. I'm not sure if it would be a bottleneck or not, but if the resources team is taking long to respond a tweet request I think it's fine to tweet first from a personal account and then retweet it.

PR #143, opened 2018-07-31
[RFC] Bootstrap the Cortex-M team

japaric:
As per RFC #136 this PR kickstarts the Cortex-M team

Note that the team composition is not final. The team is expected to grow over time. The mechanism
for joining a team is specified in RFC #136.

This is the team with most crates so it should have a large team to cope with the maintenance load.

active Cortex-M developers off the top of my head: would you be interested in helping maintain
these crates? @hannobraun @thejpster @therealprof @korken89

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,26 @@ What is it that we really want? At a broad level:
 [@pftbest]: https://github.com/pftbest
 [@thejpster]: https://github.com/thejpster
 
+### The Cortex-M team
+
+The Cortex-M team develops and maintains the core of the Cortex-M crate ecosystem.
+
+#### Members
+
+- [@japaric]
+
+#### Projects
+
+Projects maintained by this team.
+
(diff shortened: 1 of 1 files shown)


korken89:
I would indeed be very interested!

hannobraun:
I am generally interested, but my time is limited, and I'm currently prioritizing embedded-hal over the Cortex-M crates. If a certain level of activity is expected of members at all times, I should refrain from joining right now.

On the other hand, if it's okay with the other members that I won't be very active for the time being, I'd be happy to be a part of this team, and I'll pitch in as I can.

adamgreig:
Not sure if you're looking for volunteers at this stage but I'd be interested in being a part of this team and am happy to commit time to these crates.

japaric:
@korken89 @adamgreig Awesome! Thanks, you've been added to the team.

@hannobraun Thanks for considering joining. If we seem to be short staffed after the initial team is formed we'll reach out to you.

therealprof:
Count me in.

PR #145, opened 2018-08-02
move out the embedonomicon

japaric:
https://github.com/rust-embedded/embedonomicon hosts the second edition.

[Rendered second edition](https://rust-embedded.github.io/embedonomicon/)

Notables changes from the first edition which was hosted here:

- All unstable language features that are not planned for stabilization by Rust 1.31 have been
  removed and now only the stable alternatives are presented.

- The book uses `cargo-binutils` and `rust-lld` instead of `arm-none-eabi-binutils` so the later
  is not required to replicate the examples.

- All the examples can be run in QEMU. The book contains instructions on how to do so.

- The appendix covers the different levels of compiler support: how to tell if `rustc` supports a
  certain architecture, when does one need to write a target specification file and when does one
  need to use Xargo.

- The preface now list objectives, target audiences and tools required to build, run and inspect the
  examples.

r? ~@rust-embedded/resources~ @andre-richter @jamesmunns @therealprof
cc @ryankurte
@jamesmunns we'll need to do something to get the second edition on
https://embedonomicon.rust-embedded.org not quite sure what :-). Also it seems that the changes from


japaric:
Everyone has given a :+1: and @andre-richter has reviewed most of the sections at the other location so I'm going to go ahead and land this.

Issue #148, opened 2018-08-02
[RFC] Embedded Rust Community Plan


japaric:
> EDIT: Updated by @jamesmunns. See [this comment] for more details

[this comment]: https://github.com/rust-lang-nursery/embedded-wg/issues/148#issuecomment-410443841

The following items have been generally discussed over the past few months, and we would like to come up with a plan for how to support the embedded rust community moving forward. This is an initial collection, but we would love your input on the following:

* Would you like to see these components?
* What tools would make sense for these components?
* Are there any alternatives to these components that could be more valuable?
* Any other comments you would like to share

### 1. NEW: A forum

Set up at `https://forum.rust-embedded.org`. Perhaps powered by Discourse.

Meant for topics slightly larger than a single blog post or IRC communication, but slightly too ephemeral for a wiki post.

Content driven by the community

### 2. EXISTING: A Chat Room

We currently have two active chat communities:

* **IRC:**: `#rust-embedded` @ `irc.mozilla.org`.
* **Discord:**: `#wg-embedded` @ `rust-lang` discord server

Chat is meant for direct questions, random chat, technical questions.

> NOTE: This RFC does not intend to propose new or different chat systems

### 3. NEW: A blog

A static site, set up at `https://blog.rust-embedded.org`. Static site generator TBD, likely hosted on github pages, probably at `rust-embedded/blog` github space. Covers posts such as the newsletter, calls to action, and technical posts by guest contributors (they can send in posts by filing a PR to the blog project, reviewed by the `resources` team.

We should have an RSS feed.

Comments still TBD.

Content driven by the resources team and the community

### 4. NEW: A Wiki

A wiki page driven by github, probably residing in the `rust-embedded/wg` repository, which would locate it at `https://github.com/rust-embedded/wg/wiki`. We could add an alias at `https://wiki.rust-embedded.org`.

The wiki should cover things like "how does this specific chip work", or "working with these sensors", or "how to debug using a JLink debugger and GDB".

Content created and curated by the community. No guarantees on correctness.

### 5. MOVED: Our Bookshelf

These are static `mdbook` projects that will be hosted at `https://*.rust-embedded.org`. Currently we (will) have:

* The Discovery Book (learn embedded through rust)
* The Rust Embedded Book (learn embedded rust)
* The Embedonomicon (learn the low level details of embedded rust)

[...]

PR #157, opened 2018-08-10
add operational notes

japaric:
As suggested in https://github.com/rust-embedded/wg/issues/148#issuecomment-411246711.

I'm putting them in a PR for now regardless of wherever we decide to put them in the end because that seems like a better place than in my head :-).

cc @jamesmunns @adamgreig

### DIFF
--- a/ops/post-transfer.md
+++ b/ops/post-transfer.md
@@ -1,0 +1,127 @@
+# Post transfer TODO list
+
+## Travis CI
+
+Browse to https://travis-ci.org/rust-embedded/repo-name/settings and make sure that "Build pushed
+pull requests" is enabled.
+
+
+## Source
+
+Push a commit to `master` that does the following:
+
+- Creates `.github/CODEOWNERS` with the following contents:
+
+``` text
+* rust-embedded/team-name collaborator-name another-collaborator
+```
+
+- Adds a copy of [`CODE_OF_CONDUCT.md`][CoC]. Don't forget to adjust the team name and associated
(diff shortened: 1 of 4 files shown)


jamesmunns:
LGTM! Thanks for writing this all out.

ops/post-transfer.md
```diff
@@ -0,0 +1,127 @@
...
+
+Contribution to this crate is organized under the terms of the [Rust Code of
+Conduct][CoC], the maintainer of this crate, the [Cortex-M team][team], promises
+to intervene to uphold that code of conduct.
+
+[CoC]: CODE_OF_CONDUCT.md
+[team]: https://github.com/rust-embedded/wg#the-cortex-m-team
+```
+
+- If the repository uses CI, configure bors. Add a `.github/bors.toml` file with these contents:
```
adamgreig: I think this should just be `bors.toml`, not inside `.github/`?

japaric: That's where I have been putting it so far but I've been told that putting it inside `.github` also works. I think is better to make it less visible since most people (both contributors and reviewers) won't be modifying it.

adamgreig: Fine with me. Should we move it on the existing repositories then?

japaric: Yes. I also need to add `block_labels = ["needs-decision"]` to most of them.

jamesmunns:
Seems like discussion is complete, merging manually.

PR #166, opened 2018-08-12
Update README.md to propose myself

thejpster:
Added myself to HAL and Cortex-M teams as that's broadly what I was working on previously, plus the Resources because that sounds fun.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,7 @@ The Cortex-M team develops and maintains the core of the Cortex-M crate ecosyste
 - [@adamgreig]
 - [@japaric]
 - [@korken89]
+- [@thejpster]
 - [@therealprof]
 
 #### Projects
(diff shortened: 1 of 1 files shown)


jamesmunns:
Collecting votes from the following teams:

### Cortex-M

- [ ] @adamgreig 
- [ ] @korken89 
- [ ] @therealprof 
- [ ] @japaric 

### HAL team

- [ ] @hannobraun 
- [ ] @japaric 
- [ ] @therealprof 

### Resources Team

- [ ] @andre-richter 
- [ ] @jamesmunns
- [ ] @japaric 
- [ ] @therealprof

therealprof:
@thejpster Great to have you on board!

jamesmunns:
If/when @thejpster is added to the `/all` organization, we should pull his name from https://github.com/rust-embedded/wg/blob/master/.github/CODEOWNERS

japaric:
The teams have been updated.

PR #181, opened 2018-08-18
Update Embedded Book for mdbook 0.2.0

Havvy:
mdbook 0.2.0's has a breaking change such that links are relative to the file they are in instead of the top level of the book. Luckily, this is your only link like that. So once you accept this PR you can build with mdbook 0.2.0 which is what we're trying to switch to for the the bookshelf.

### DIFF
--- a/books/embedded-rust-book/src/intro/hardware.md
+++ b/books/embedded-rust-book/src/intro/hardware.md
@@ -12,7 +12,7 @@ Let's get familiar with the hardware we'll be working with.
 ## STM32F3DISCOVERY (the "F3")
 
 <p align="center">
-<img title="F3" src="assets/f3.jpg">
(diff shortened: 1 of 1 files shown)


hannobraun:
cc @rust-embedded/resources

jamesmunns:
Sorry for the delay, I'm still in the US running around after RustFest. I should hopefully have some time this week to review and address this. Thanks for the contribution @Havvy!

japaric:
Thanks for the PR, @Havvy. The book was moved into its own repository (see https://github.com/rust-embedded/book) just a while ago but we forgot the remove the old version from this repository. Sorry for the confusion.

 Also, the problem with mdbook 0.2 and links has already been fixed in rust-embedded/book#13.

PR #186, opened 2018-08-21
Propose myself for the cortex-m & hal teams

ithinuel:
I apply to these teams because that plus rust is all I like to do !


thejpster:
+1

adamgreig:
We have a majority approval from both teams.

Welcome @ithinuel!

bors r+

japaric:
@ithinuel I have sent you the invitation. You should have received an e-mail. You can also just browse to https://github.com/rust-embedded and you'll see the invitation, I think.

Issue #147, opened 2018-08-02
[RFC] Curated list of projects


japaric:
We have a curated list of crates and tools in [awesome-embedded-rust] but there's an opportunity for
also showing embedded Rust projects in a more compelling manner than just a bullet list.

[awesome-embedded-rust]: https://github.com/rust-embedded/awesome-embedded-rust

Let's make a web page that list projects with pictures and videos! And link it from the embedded WG
web page that will be on rust-lang.org

Here's a detailed strawman proposal to get the conversation started

### Repository name

"Made with embedded Rust" OR "Awesome embedded Rust projects"

### Project submission requirements

- Obviously, it must involve hardware that runs Rust code.
- Must compile on stable by Rust 1.31.
- The code must be public.
  - I feel this is very useful for developers that want to see how an embedded Rust application is
    structured

### Submission format

- Project name
- Author name
- Project website (+ link to repo) OR project repository
- Project description
- Image, GIF or video of the hardware

### Unresolved questions

At what point do we consider something a "project"? Do we consider the examples of a board support
crate as a "project" that can be submitted here? Do we consider a book a "project" too?

Should we limit submissions to projects that are "done"?

Thoughts?

cc ~@rust-embedded/resources~ @andre-richter @jamesmunns @therealprof
cc @cr1901 @thejpster

little-arhat:
Why "Must compile on stable by Rust 1.31" requirement for _projects_ (as opposed to tools/libs)?

japaric:
> Why "Must compile on stable by Rust 1.31" requirement for _projects_ (as opposed to tools/libs)?

Because they will be showcased as *examples* to follow or to use as reference -- people *will* read
the source code. As such the projects should showcase good programming practices and at a minimum we
should show that one does not need to rely on something that will break tomorrow to write embedded
Rust applications.

Furthermore, I'm pretty sure plenty of people will try to build the projects for several reasons:
because they heard that Rust makes cross compilation is easy and that the build system is great, or
perhaps because they want to check that the binary artifacts are small. Whatever the reason, it
would be pretty bad if they try and the project fails build on stable.

On the other hand, most people won't be reading the source code of the tools and libraries they use.
Tools don't need to compile on stable to be used with a stable toolchain. And if a library doesn't
compile on stable you can look for an alternative on crates.io.

Of course, this is my opinion. If the majority thinks that "compiles on  stable" should not be a
requirement then I'll concede.

[...]

PR #187, opened 2018-08-21
[RFC] Embedded Rust Showcase

adamgreig:
Follow-up from #147 and today's IRC meeting.

[Rendered](https://github.com/rust-embedded/wg/blob/embedded-rust-showcase/rfcs/0000-embedded-rust-showcase.md)

cc @rust-embedded/resources

### DIFF
--- a/rfcs/0000-embedded-rust-showcase.md
+++ b/rfcs/0000-embedded-rust-showcase.md
@@ -21,7 +21,7 @@ developers.
 The main embedded Rust website will contain a shortlist of especially notable
 projects, but is aimed at presenting a very high-level overview to all users
 including non-technical users. This proposed list would be able to present more
-projects in more detail, and will require that their source code is publically
+projects in more detail, and will require that their source code is publicly
 available so embedded developers interested in Rust can inspect it.
 
 # Detailed design
@@ -50,19 +50,27 @@ To be considered for inclusion, projects must:
 
 * Involve embedded Rust, in other words, use Rust and run on embedded hardware
 
-* Have publically available source code
+* Have publicly available source code
 
 The objective of this list is to showcase Rust code in action; we can't do this
 if people can't read the code! There might be space on the main Embedded Rust
 website to showcase projects known to use Rust but without public code. We
-don't require that the source code is under an open source license; just that
-it's available for interested users to read.
+don't require any specific license; just that it's available for interested
+users to read.
 
 * Have working CI builds
 
 Since we want people to be able to learn from the code, we require it at least
 builds successfully. Working CI also shows what versions of Rust it builds on,
 which is useful to establish if a project works on stable Rust.
+
+* Have build instructions
+
+These might be as simple as `cargo build` or might document any specific
+Rust versions, build oddities, or other steps required to produce the
+final firmware. Projects could also document their physical build, for example
+whether a particular development board was used, or what custom hardware is
+present, but this is not required.
 
 * Have at least one photograph/video/GIF of the project in action
 
--- a/rfcs/0000-embedded-rust-showcase.md
+++ b/rfcs/0000-embedded-rust-showcase.md
@@ -1,0 +1,101 @@
+# Summary
+[summary]: #summary
+
+This RFC proposes establishing a new repository for a web page advertising
+noteworthy embedded Rust projects. This new resource would go beyond the
+existing [awesome-embedded-rust] list by allowing long form
+descriptions, photographs, and videos to best demonstrate the projects.
+
+[awesome-embedded-rust]: https://github.com/rust-embedded/awesome-embedded-rust
+
+# Motivation
+[motivation]: #motivation
+
+We would like a well-presented list of interesting embedded projects using Rust
+to advertise Rust's abilities in this domain. Existing lists are mainly
+targeted at embedded Rust developers looking for libraries; this list would
+instead be advertising the final projects people have accomplished with Rust,
+and so should appeal to potential users who are not already embedded Rust
+developers.
+
+The main embedded Rust website will contain a shortlist of especially notable
+projects, but is aimed at presenting a very high-level overview to all users
+including non-technical users. This proposed list would be able to present more
+projects in more detail, and will require that their source code is publically
+available so embedded developers interested in Rust can inspect it.
+
+# Detailed design
+[design]: #detailed-design
+
+## New Repository
+
+We create a new repository, `embedded-rust-showcase` (TBC), in the
+rust-embedded organisation and under the existing Resources team. It will use
+the same underlying technology as the upcoming embedded Rust website to render
+Markdown to a web page.
+
+## New Web Page
+
+The new web page would live at https://showcase.embedded-rust.org (TBC).
+
+Each project is listed with a photograph/GIF/video, description, and other
+details (see Submission section).
+
+Additionally, each project has a badge indicating whether it builds on stable
+Rust or requires a nightly compiler.
+
+## Project Requirements
+
+To be considered for inclusion, projects must:
+
+* Involve embedded Rust, in other words, use Rust and run on embedded hardware
+
+* Have publically available source code
+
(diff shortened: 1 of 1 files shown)


rfcs/0000-embedded-rust-showcase.md
```diff
@@ -0,0 +1,101 @@
...
+including non-technical users. This proposed list would be able to present more
+projects in more detail, and will require that their source code is publically
+available so embedded developers interested in Rust can inspect it.
+
+# Detailed design
+[design]: #detailed-design
+
+## New Repository
+
+We create a new repository, `embedded-rust-showcase` (TBC), in the
```
hannobraun: Can be shortened to just `showcase`, due to the org namespace: `rust-embedded/showcase`

japaric:
Thanks for writing this, @adamgreig!

This proposal (PR) needs at least 10 votes (approvals) from @rust-embedded/all
to be accepted. After the proposal has been accepted we'll proceed to implement
it.

@rust-embedded/all please vote on this proposal using [pull request reviews].
Or if you have a concern leave a comment.

rfcs/0000-embedded-rust-showcase.md
```diff
@@ -0,0 +1,101 @@
...
+rust-embedded organisation and under the existing Resources team. It will use
+the same underlying technology as the upcoming embedded Rust website to render
+Markdown to a web page.
+
+## New Web Page
+
+The new web page would live at https://showcase.embedded-rust.org (TBC).
```
hannobraun: You mean http://www.rust-embedded.org/? embedded-rust.org ic rurrently unregistered.

rfcs/0000-embedded-rust-showcase.md
```diff
@@ -0,0 +1,101 @@
...
+
+* Project name
+* Author name
+* Project website/repository
+* Project description
+* Images, GIFs, or video of the project
+* Whether or not the project builds on stable Rust
+
+After a project is submitted, the resources team will review the PR and
+vote on inclusion. They may consult other relevant teams in the process,
```
hannobraun: Maybe a full vote is overkill. This isn't some highly sensitive decision. If one or two team members think something should be included, I think that's good enough.

rfcs/0000-embedded-rust-showcase.md
```diff
@@ -0,0 +1,101 @@
...
+but final approval of submissions is at the discretion of the resources team.
+
+# Alternatives
+
+* Instead of badges for builds-on-stable projects, the nightly-only projects
+  could be binned together at the bottom of the page.
+* The web page could instead be maintained as part of the main embedded-rust
+  website, existing as a page in that website.
+* We could extend awesome-embedded-rust to permit longer descriptions of
+  projects.
```
hannobraun: I think it would be better to keep the showcase separate and focus awesome-embedded-rust on libraries, tools, etc.

rfcs/0000-embedded-rust-showcase.md
```diff
@@ -0,0 +1,101 @@
...
+* We could integrate this concept into the existing awesome-embedded-rust
+  repository, but still render the projects to a web page.
+
+# Unresolved Questions
+[unresolved]: #unresolved
+
+* Final name for repository/web page
+* The specific technical solution for rendering the repository contents to a
+  web page
```
hannobraun: I don't think that last point needs to be part of the RFCs. As long as we agree on the general idea (i.e. static website generator), whoever ends up implementing it is most qualified to decide whether to use Jekyll, Gutenberg, or whatever else.

adamgreig:
Thanks for the comments @hannobraun! I think they're worth including and I'll update the RFC with them this evening.

adamgreig:
I had some time so addressed your comments in 4177a15 (though this has dismissed your review).

rfcs/0000-embedded-rust-showcase.md
```diff
@@ -0,0 +1,99 @@
...
+
+To be considered for inclusion, projects must:
+
+* Involve embedded Rust, in other words, use Rust and run on embedded hardware
+
+* Have publically available source code
+
+The objective of this list is to showcase Rust code in action; we can't do this
+if people can't read the code! There might be space on the main Embedded Rust
+website to showcase projects known to use Rust but without public code. We
```
therealprof: To me if the code is available to the public it is very much the meaning of the term **open source**. I think what you're trying to say is that the specific choice of license doesn't matter as long as the source is available. Not sure how to properly word that, though.

rfcs/0000-embedded-rust-showcase.md
```diff
@@ -0,0 +1,99 @@
...
+it's available for interested users to read.
+
+* Have working CI builds
+
+Since we want people to be able to learn from the code, we require it at least
+builds successfully. Working CI also shows what versions of Rust it builds on,
+which is useful to establish if a project works on stable Rust.
+
+* Have at least one photograph/video/GIF of the project in action
+
```
therealprof: I would like to make build/configuration/installation instructions at least a soft requirement so that interested parties can reproduce the work or at least have a rough guideline for their own projects.

therealprof:
@adamgreig Thanks for writing this up, this looks very good to me.

rfcs/0000-embedded-rust-showcase.md
adamgreig: I don't understand what change you'd like: the RFC already says the code must be publicly available but that it doesn't require an open source license.

rfcs/0000-embedded-rust-showcase.md
adamgreig: How should we specify a soft requirement? I was thinking that if CI is working that means it should be pretty easy to figure out how to build the project, but since this showcase is about complete projects I don't imagine many people will be reproducing the whole project anyway (you'd presumably need custom hardware); instead we want to show off what projects are possible.

japaric: I will be one of those curious people that will compile the project just to see how big the binary is. I might also look at some parts of the disassembly, in particular interrupt handlers, if it seems that the program contains time sensitive / real time portions. So I would give extra points for build instructions and double extra points if the CI includes the output of `size`.

adamgreig: That's fair. Do you think having working CI is enough of a requirement or do we want to require specific build instructions too?

rfcs/0000-embedded-rust-showcase.md
therealprof: Not sure, just reads odd to me. It kind of implies open source in the definition of the OSI which many people do not agree with. Maybe shorten the sentence to: "We don't require any specific license; just that
it's publicly available for interested users to read."?

rfcs/0000-embedded-rust-showcase.md
japaric: I will first try `cargo build --release` (on w/e rustc channel I happen to be on) and then check the CI builds if that doesn't work but I'm sure most people will give up after the first error so I would say include build instructions if it's more complex than doing `cargo build --release` from the root of the repo -- assuming that the user will have stable as their default channel. If nightly is required they can just put a `rustc-version` file in the root of the repository and not write build instructions.

therealprof: Well, the project may or may not require custom hardware but with embedded projects there're more steps involved than just "fire up the compiler, then run it". It may require custom hardware or run on a development board. It's likely to require additional hardware, then it would be interesting how that is hooked up. Even if an interested reader is not capable or willing to reproduce it it might spark ideas for new projects or animate to tinker which I think would be a nice outcome of the showcase display.

adamgreig:
I've updated to address @therealprof 's comments.

japaric:
This RFC is now officially approved so I'm going to merge it.

bors r+

@rust-embedded/resources anyone up for implementing this? We used [Gutenberg](https://github.com/Keats/gutenberg) for the blog, perhaps there's some template we could use here?

Issue #156, opened 2018-08-10
Website: update DNS infrastructure


nastevens:
Currently rust-embedded.org (and redirect sites rust-embedded.com and areweembeddedyet.com) have DNS managed directly by 1and1.com's DNS tools. Those tools are frankly garbage, and don't allow even basic things like multiple IP addresses on an A record.

I am planning to move DNS management to AWS Route 53 and implement the following proposed layout for `rust-embedded.org`:
  `@` -> `A` record pointed to Github apex IP addresses (https://help.github.com/articles/setting-up-an-apex-domain/)
  `www` -> `CNAME` pointed to `rust-embedded.org`, which will automatically redirect to https://github.com/rust-embedded/rust-embedded-www
  `embedonomicon` -> `CNAME` pointed to `rust-embedded.org`, which will automatically redirect to https://github.com/rust-embedded/embedonomicon
  `book` -> `A`/`AAAA` records pointed at @jamesmunns's server at 46.101.143.249/2a03:b0c0:3:d0:0:0:cb7:5001

`rust-embedded.com` will redirect all subdomains to their equivalent at `rust-embedded.org`. `areweembeddedyet.com` redirects to the main page for the apex domain and `www`, but doesn't redirect any of the subdomains (book, embedonomicon).

Comments welcome - I will set up the AWS records first so we can try them out before cutting over the registrar to use those addresses.

CC @jamesmunns @ryankurte 
Refer to https://github.com/rust-embedded/embedonomicon/pull/12

ryankurte:
I don't know what the route53 tools are like for sharing, but cloudflare supports organisational accounts (and does everything we would need + some extra goodies) for free, which our one day ops team might find useful. I'm sure the same can be achieved with route53, but I'm not sure how difficult it might be?

jamesmunns:
Sounds good to me @nastevens, thanks for opening up this issue. I agree with @ryankurte, it would be nice not to have a single point of failure/put all the burden on you whenever we need to make changes.

jamesmunns:
Another option would be to transfer ownership of the domain to the Rust organization itself, and manage our DNS through whatever process `rust-lang.org` uses. I'm not too familiar with this, or if this is a good or bad idea. @japaric might know though.

nastevens:
[...]

Issue #153, opened 2018-08-07
More communication channels with the teams


japaric:
We now have 5 brand new teams :tada:. That's awesome but I think we need to make it clearer how someone can contact the teams.

Ideas:

- Add your IRC handle to the README. (Easy)
- Add your discord handle to the README. (Easy)
- Set up e-mail aliases for each team (e.g. `cortex-m@rust-embedded.org`). This private communication channel is required to receive reports regarding moderation and violations of the code of conduct. (Not sure what would be needed to set up those)

A perhaps orthogonal question is how *the team* will notify the stakeholders (e.g. users of the Cortex-M crates) when their input is required to make decisions on breaking changes and other RFCs (e.g. rust-embedded/cortex-m#82)? I can't think of a good solution for this. I try to cc stakeholders when I open an RFC but I would only reach out to more or less active contributors (also, my memory is not perfect).

The rust-lang org tackles this problems in the following ways: they have a dedicated RFC repository which you can subscribe to (IMO, traffic is too high and the topics so varied that this will overwhelm most people); they post new RFCs and which RFCs are entering FCP  in their weekly newsletter (their newsletter has a much larger audience than ours so this may not be as effective; also our newsletter has a biweekly cadence).

Thoughts?

cc @rust-embedded/all

ryankurte:
Maybe we could have join links for the discord and IRC channels in our documentation too?

I see to always end up using mailgun with a forwarding rule to setup / manage email aliases, it's far from elegant, but works if noone has a better solution (and could perhaps be a step towards email subscription to the newsletter too).

WRT stakeholders, it's probably not going to help everyone, but perhaps we could we also render out the RFCs as a static site w/ RSS feed (as well as posting them in the newsletter)?
Then it's pretty easy to automate into whatever tooling / chat clients / anything else people are using, and doesn't require reading the newsletter to keep up / isn't effected by the newsletter cadence.

japaric:
Nominating for discussion in the next meeting (#150)

japaric:
@andre-richter The discord server is an ["experiment"](https://internals.rust-lang.org/t/exploring-new-communication-channels/7859) by the Rust core team. We have a channel over there because they created one for each domain WG (without asking any of us AFAIK).

> Will we have a preference between both?

I haven't seen any current member of the WG (other than me -- I check the channel sparingly) use the embedded-wg discord channel. And I have received one comment (on IRC) about *not* wanting to switch the meetings to discord.

[...]

PR #189, opened 2018-08-22
add e-mail addresses for the teams

japaric:
This a proposal to give each team an e-mail address. As discussed in the last
meeting we'll be using the `team-name@teams.rust-embedded.org` format for the
e-mail addresses.

This proposal (PR) needs at least 10 votes (approvals) from @rust-embedded/all
to be accepted. After the proposal has been accepted we'll proceed to implement
it.

@rust-embedded/all please vote on this proposal using [pull request reviews].
Or if you have a concern leave a comment.

[pull request reviews]: https://help.github.com/articles/about-pull-request-reviews/

### DIFF
--- a/README.md
+++ b/README.md
@@ -353,6 +353,16 @@ most of them are members of one of the teams listed above.
 
 ### Contact
 
+Each team can be privately contacted via the following e-mail addresses:
+
+- cortex-m@teams.rust-embedded.org
+- embedded-linux@teams.rust-embedded.org
+- hal@teams.rust-embedded.org
+- msp430@teams.rust-embedded.org
+- riscv@teams.rust-embedded.org
+- resources@teams.rust-embedded.org
+- tools@teams.rust-embedded.org
+
 You can usually find the members of the embedded WG on the #rust-embedded channel (server:
 irc.mozilla.org).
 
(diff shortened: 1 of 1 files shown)


japaric:
10 people have approved this proposal so I'm going to mark it as "accepted".

Before we land this PR we need to implement the e-mail aliases but I believe that's blocked on updating the DNS infrastructure (#156) so I'm going to mark this as blocked for now. cc @nastevens

nastevens:
Initial changes are up for DNS infrastructure (https://github.com/rust-embedded/wg/issues/156) so we can start thinking about this again. I have not personally set up mailing list infrastructure before - is there anyone with experience in this that can work with me?

japaric:
@ryankurte mentioned in #153 that they have experience setting up mail aliases using mailgun.

ryankurte:
Yep! The mailgun add-domain stuff should get you most of the way there, then I just add routes for each address. Can check your work here or via email if you would like, and let me know if you need any more help?

nastevens:
@ryankurte - If you're able to take a look at #156 that gets the base DNS terraform config set up. From there it should be straightforward to add the `teams` subdomain and appropriate MX records to make the lists happen.

japaric:
@rust-embedded/infrastructure now that the DNS stuff has been sorted out, any chance you could look into setting up the e-mail aliases?

japaric:
The infra team is currently setting up the team e-mails!

I have updated this PR to list only the team e-mails that are in place. That way we can land this PR now and then add the other e-mails in follow-up PRs.

r? anyone

PR #212, opened 2018-09-25
README: point readers to the embedded Rust book and the bookshelf

japaric:
r? @rust-embedded/all (anyone)

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,12 @@ a great choice for embedded development.
 
 [issue tracker]: https://github.com/rust-lang-nursery/embedded-wg/issues
 
+**Want to get started with embedded development with Rust?** Check out our
+[embedded Rust book][book] and the rest of our [bookshelf].
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -9,6 +9,12 @@ a great choice for embedded development.
 
 [issue tracker]: https://github.com/rust-lang-nursery/embedded-wg/issues
 
+**Want to get started with embedded development with Rust?** Check out our
+[embedded Rust book][book] and the rest of our [bookshelf].
```
therealprof: Uhm, clicking that link leads to a page with a TODO right at the beginning. I think we should fix this ASAP to not leave a bad first impression.

jamesmunns: As for context, I was borrowing from the [Embedded Rust Book Introduction](https://doc.rust-lang.org/book/2018-edition/ch00-00-introduction.html), and I am okay to just pull it.

Let me know if you want me to open up the PR

PR #222, opened 2018-09-29
Transfer gpio-cdev repo ownership to embedded Linux team

posborne:
Currently hosted at https://github.com/posborne/rust-gpio-cdev.  The repo would be migrated to rust-embedded/gpio-cdev.

[`gpio-cdev`](https://crates.io/crates/gpio-cdev) is a new crate for interacting with GPIOs in Linux using the character device API (the sysfs interface is now deprecated and will be going away in future kernels [2020]).  Differences from the old API are [covered in the project README](https://github.com/posborne/rust-gpio-cdev#sysfs-gpio-vs-gpio-character-device).  Now that the initial work for the crate is done and an initial version of this crate, I want to transfer ownership over to the working group.

I didn't see a documented process for this kind of transfer, so I'm winging it a bit and using the change to the team ownership as a review checkpoint for the overall inclusion decision.  In this case, several (but not all) of the changes from the [Post Transfer TODO List](https://github.com/rust-embedded/wg/blob/master/ops/post-transfer.md) have already been done on the repo.  The other would be done once moved over -- several of the other Linux repos which have been around for awhile need a refresh along the same lines (many of those date back to when I created this GH org a few years back).

r? @rust-embedded/embedded-linux 

---

Since this crate is very new (0.1 was just published today), review and feedback on the APIs is, of course, greatly appreciated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,6 +181,7 @@ The embedded Linux team develops and maintains the core of the embedded Linux cr
 
 Projects maintained by the embedded Linux team
 
+- [`gpio-cdev`]
 - [`gpio-utils`]
 - [`i2cdev`]
 - [`linux-embedded-hal`]
@@ -376,6 +377,7 @@ https://mozilla.logbot.info/rust-embedded
 [`cortex-m`]: https://github.com/rust-embedded/cortex-m
 [`cross`]: https://github.com/rust-embedded/cross
 [`embedded-hal`]: https://github.com/rust-embedded/embedded-hal
+[`gpio-cdev`]: https://github.com/rust-embedded/gpio-cdev
 [`gpio-utils`]: https://github.com/rust-embedded/gpio-utils
 [`i2cdev`]: https://github.com/rust-embedded/rust-i2cdev
 [`itm`]: https://github.com/rust-embedded/itm


thejpster:
I had a look - this all seems pretty neat and tidy.

One question - is it possible to implement the `embedded-hal` traits against this? That would give Linux users access to a bunch of GPIO-based drivers.

posborne:
> One question - is it possible to implement the embedded-hal traits against this? That would give Linux users access to a bunch of GPIO-based drivers.

Adding implementations of both futures and the hal traits is definitely something that I think we probably should look to do with future releases.

Looking at linux-embedded-hal and determining if it would make sense to push the implementation of traits to the crates rather than requiring the intermediate crate with wrapper types is something I need to look at a little bit more.

RandomInsano:
One thing, rstr_lcpy() seems non-Unicode safe here. An `str` can have one to four bytes while a `char` has one.

Is the assumption that the consumer_label will never have non-ASCII text in it? Digging into this via cell phone, I wasn’t able to see if the Linux interface itself is Unicode aware.

posborne:
> One thing, rstr_lcpy() seems non-Unicode safe here. An str can have one to four bytes while a char has one.

It may not have the desired affect but isn't unsafe.  You'll just end up with a string in the kernel that when read back cannot be interpreted as a valid unicode string.

> Passing in a CStr would make this more explicit.

Ergonomically, I don't care for that approach.  I might consider adding an error if truncation on the input `&str` would fail.  For what it is, documenting that it should just be ascii might be fine as well so long as it doesn't buffer overflow or anything.

RandomInsano:
Ah, today I learned &str.len() is in bytes:
https://doc.rust-lang.org/std/primitive.str.html#method.len

posborne:
> Ah, today I learned &str.len() is in bytes: https://doc.rust-lang.org/std/primitive.str.html#method.len

Indeed; makes some sense given that String/str are, at some level, just Vec/slice and the length in bytes is the only thing you have available without any significant `O(n)` cost on call (or tracking overhead for all other operations).

PR #223, opened 2018-09-30
Rename ops team to infrastructure team

ryankurte:
Ambiguity between working group operations and `ops` / `operations` team was raised at the last meeting. This PR renames the `ops` team to the `infrastructure` team to remove that ambiguity, as suggested in https://github.com/rust-embedded/wg/issues/217#issuecomment-425669425.

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,9 +227,9 @@ Projects maintained by this team
 - [`msp430-rt`]
 - [`msp430`]
 
-### The ops team
+### The infrastructure team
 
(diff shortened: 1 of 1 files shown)


nastevens:
Can this RFC be merged per changes in #206? There are no outstanding issues and we're well over 33%.

japaric:
@nastevens this PR only needs majority from the team and no concerns from the other WG members so this is good to go!

bors r+

PR #207, opened 2018-09-09
[RFC] Initial Cortex-A team

andre-richter:
This is a proposal for creating a Cortex-A team.

This is based on and will close issue #182. Please refrain from approving until I removed the [DRAFT] from the title and every proposed member has his points on the agenda.

@raw-bin, @wizofe and me will be the initial members.

The following work is planned for this team:

### Maintain a set of ecosystem crates under the embedded WG

- [ ] Write a tool that generates accessors for the ARM System Register XML.
- [ ] Move [cortex-a](https://github.com/andre-richter/cortex-a) to the WG and keep extending/completing it.
    - [ ] If feasible, feed the XML accessors into the crate.
    - [ ] Add ready-to-use idioms for bare-metal use cases.
        - [ ] MMU idioms
        - [ ] GIC idioms
        - [ ] ...? 
- [ ] Move [register-rs](https://github.com/rust-osdev/register-rs) to the WG. CC @phil-opp for sign-off.
- [ ] Discuss if something the likes of `cortex-m-rt` is feasible for cortex-a.

### Provide resources for learning

- [ ] Move [rust-raspi3-tutorial](https://github.com/andre-richter/rust-raspi3-tutorial) to the WG

### Make it possible to build AArch64 kernels / programs on stable.

- [ ] Add a `rust-std` component for `aarch64-unknown-none`
- [ ] Differentiate into further aarch64 targets, e.g. a non-fp version of `aarch64-unknown-none`, and add them as well.

### Stabilization

- [ ] Push for a stable core::arch::arm module (see #63) to make the `cortex-a` crate (see its asm module) work on stable w/o depending on GCC.

------------------

@raw-bin and @wizofe, kindly let me know your agenda points and I will edit them into this comment.
I understand your Redox work will most probably be a part of it, but I want to leave it to you guys to describe it.

Once we are set, I will remove the [DRAFT] and open this for members to approve.

@jamesmunns I tend to agree to your thought about having a microprocessor group. However, I think that cortex-a alone is such a complex architecture, it can easily keep the initial three of us busy for years to come.
I would argue that we would need separate teams for PPC, RISC-V and the likes.

PR #227, opened 2018-10-09
Add register-rs to the project list of new Cortex-A team

phil-opp:
The crate is currently hosted under the rust-osdev organization: https://github.com/rust-osdev/register-rs. Moving it to the WG was part of the recently accepted [RFC for a Cortex-A team](https://github.com/rust-embedded/wg/pull/207#issuecomment-427606754).

This PR follows the process outlined in https://github.com/rust-embedded/wg/blob/master/rfcs/0136-teams.md#adopting-projects.

cc @andre-richter @japaric

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,6 +144,7 @@ The Cortex-A team develops and maintains the core of the Cortex-A crate ecosyste
 
 Projects maintained by this team.
 
+- [`register-rs`]
 - To be updated soon.
 
 ### The Cortex-M team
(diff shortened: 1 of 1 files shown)


andre-richter:
Thanks phil!

@japaric: Since moving the crate was part of the referenced and accepted RFC already, we can skip the voting majority this time, right?

japaric:
@andre-richter you only need voting majority from the Cortex-A team (and no concerns from other WG members) in this case, but seems fine to skip the voting in this case.

andre-richter:
@phil-opp I just checked and don't have the rights to transfer ownership. Please take it away :)

phil-opp:
@andre-richter I don't have the rights to create repositories under rust-embedded, so I can't do the transfer either. But I just adjusted the organization rights so that you should be able to do it now.

andre-richter:
Thanks again @phil-opp.

The move has happened.

bors r+

PR #287, opened 2019-01-06
add cortex-r, dcc and panic-dcc to crates maintained by the cortex-r team

japaric:
DCC (Debug Communication Channel) is a debug interface similar to Cortex-M's
ITM.

r? @rust-embedded/cortex-r

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,8 +185,8 @@ The Cortex-R team develops and maintains the core of the Cortex-R crate ecosyste
 
 #### Projects
 
+- [`arm-dcc`]
 - [`cortex-r`]
-- [`dcc`]
 - [`panic-dcc`]
 
 ### The embedded Linux team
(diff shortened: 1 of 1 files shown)


thejpster:
Oh. Where I'm from, DCC is the official standard for digitally controlling model trains. Can we call this `cortex-dcc` or `arm-dcc` or `debug-comm-channel`, or something?

paoloteti:
**DCC = Dual-Clock Comparator** on lockstep ARM devices. So, even if DCC is vendor specific and not from ARM, `arm-dcc` may be is not the best name too.

eddyp:
But this code would support the arm/cortex implementation of dcc, so it should be OK, right?
BTW, I consider it a design flaw that cargo has a flat name space since name collisions can occur really easily. For instance, there is a git tool that predates Linus' vcs...

japaric:
Renamed dcc to arm-dcc in the README.

@paoloteti you'll have to approve this PR again.

japaric:
sudo
bors r=paoloteti

japaric:
The arm-dcc repository has been moved into the org: https://github.com/rust-embedded/arm-dcc

PR #304, opened 2019-02-05
MSRV policy

### LINKED ISSUE #285: Minimum supported Rust version policy?
Now that embedded Rust has reached the stable channel I think we should consider setting an org-level "minimum supported Rust version" policy. By "minimum supported Rust version" I actually mean "how many stable Rust versions should the latest minor version of a (pre-1.0) crate support?".

For example, we could support only the latest stable release; meaning that once a new stable version of Rust, e.g. 1.32, is out we can issue a new minor version of e.g. `cortex-m` that no longer compiles with Rust 1.31 and requires >=1.32 to build. Note that this doesn't necessarily mean that we'll (artificially) bump the required Rust version whenever we release a new minor version; however, the user should be aware that moving to a new minor version of one our crates may require a compiler update if they are using an old stable compiler.

I haven't seen a common trend like "we only guarantee that the three latest stable Rust releases are supported by our latest crate release" in the ecosystem but I have seen some crates, like [`syn`], that go great lengths to support very old compilers (e.g. `syn` supports Rust >=1.15). We could do something like that as well.

[`syn`]: https://github.com/dtolnay/syn#proc-macro-shim

Thoughts? Should we provide (and document) any guarantee about the Rust versions a crate support? Should we try to make that guarantee uniform across all our crates? What would be a reasonable policy? Last 3 releases (one nightly -> beta -> stable cycle)? 2 (one quarter)? always >=1.31 (>= 2018 edition)?

japaric:
closes #285

### DIFF
--- a/ops/msrv.md
+++ b/ops/msrv.md
@@ -1,0 +1,49 @@
+# Minimum Supported Rust Version (MSRV)
+
+This text documents the MSRV policy used in the crates maintained by the WG.
+
+- The MSRV shall never be greater than the second latest minor stable release at
+  the moment the crate is published. For example, if the latest stable Rust
+  version was 1.32.1 a new minor release of the `cortex-m` crate can't bump the
+  MSRV higher than 1.31.0.
+
+- Changing the MSRV of a crate is a breaking change and requires a semver bump:
+  minor version bump if the crate is pre-1.0 and a major version bump if the
+  crate is (post-)1.0.
+
+- Cargo features are allowed to depend a on Rust version greater than the MSRV,
+  even a nightly compiler. For example, a "const-fn" can bump the required
+  version to 1.34.0 (effectively, nightly channel at the time of writing).
+
+- When doing a semver bump (i.e. minor version bump in pre-1.0 crates), for
+  whatever reason, the team in charge of the crate should consider bumping the
+  MSRV as well.
+
+- The MSRV will be documented in the crate level documentation of the crate,
+  like so:
+
+``` markdown
+# MSRV
+
+This crate is guaranteed to compile on stable Rust 1.31.0 and up. It *might*
+compile with older versions but that may change in any new patch release.
+```
+
+- The MSRV will be tested in the CI of the crate as a separate build job, like
+  so:
+
+``` yaml
+matrix:
+  include:
+      # MSRV
+    - env: TARGET=thumbv7m-none-eabi
+      rust: 1.31.0
+
+    - env: TARGET=thumbv7m-none-eabi
+      rust: stable
+
+    - env: TARGET=thumbv7m-none-eabi
+      rust: beta
+
+      # etc.
+```
--- a/ops/msrv.md
+++ b/ops/msrv.md
@@ -2,10 +2,10 @@
 
 This text documents the MSRV policy used in the crates maintained by the WG.
 
-- The MSRV shall never be greater than the second latest minor stable release at
-  the moment the crate is published. For example, if the latest stable Rust
-  version was 1.32.1 a new minor release of the `cortex-m` crate can't bump the
-  MSRV higher than 1.31.0.
+- The MSRV shall always be strictly smaller than the latest available minor
+  stable release. For example, if the latest stable Rust version was 1.32.1 a
+  new minor release of the `cortex-m` crate can't bump the MSRV higher than
+  1.31.0.
 
 - Changing the MSRV of a crate is a breaking change and requires a semver bump:
   minor version bump if the crate is pre-1.0 and a major version bump if the
--- a/ops/msrv.md
+++ b/ops/msrv.md
@@ -23,7 +23,7 @@ This text documents the MSRV policy used in the crates maintained by the WG.
   like so:
 
 ``` markdown
-# MSRV
+# Minimum Supported Rust Version (MSRV)
 
 This crate is guaranteed to compile on stable Rust 1.31.0 and up. It *might*
 compile with older versions but that may change in any new patch release.
--- a/ops/msrv.md
+++ b/ops/msrv.md
@@ -5,7 +5,7 @@ This text documents the MSRV policy used in the crates maintained by the WG.
 - The MSRV shall always be strictly smaller than the latest available minor
   stable release. For example, if the latest stable Rust version was 1.32.1 a
   new minor release of the `cortex-m` crate can't bump the MSRV higher than
-  1.31.0.
+  1.31.
 
 - Changing the MSRV of a crate is a breaking change and requires a semver bump:
   minor version bump if the crate is pre-1.0 and a major version bump if the
@@ -25,7 +25,7 @@ This text documents the MSRV policy used in the crates maintained by the WG.
 ``` markdown
 # Minimum Supported Rust Version (MSRV)
 
-This crate is guaranteed to compile on stable Rust 1.31.0 and up. It *might*
+This crate is guaranteed to compile on stable Rust 1.31 and up. It *might*
 compile with older versions but that may change in any new patch release.
 ```
 


ops/msrv.md
```diff
@@ -0,0 +1,49 @@
...
+  the moment the crate is published. For example, if the latest stable Rust
+  version was 1.32.1 a new minor release of the `cortex-m` crate can't bump the
+  MSRV higher than 1.31.0.
+
+- Changing the MSRV of a crate is a breaking change and requires a semver bump:
+  minor version bump if the crate is pre-1.0 and a major version bump if the
+  crate is (post-)1.0.
+
+- Cargo features are allowed to depend a on Rust version greater than the MSRV,
+  even a nightly compiler. For example, a "const-fn" can bump the required
```
eddyp: I think this is a little confusing. I suspect the intention is to allow something like configs which include code that depends on nightly features, since those configs are not necessarily available for stable. Is this correct?

If so, I think since rephrasing is needed.

ops/msrv.md
```diff
@@ -0,0 +1,49 @@
+# Minimum Supported Rust Version (MSRV)
+
+This text documents the MSRV policy used in the crates maintained by the WG.
+
+- The MSRV shall never be greater than the second latest minor stable release at
```
eddyp: I think it's simpler and clearer if we say "shall always be strictly smaller than the latest available stable release"

korken89:
I agree with @eddyp comments, else it LGTM

ops/msrv.md
adamgreig: What do you mean by 'configs' exactly? I think as-written it's OK, it's talking specifically about [features](https://doc.rust-lang.org/cargo/reference/manifest.html#the-features-section) which we'd allow to require more recent Rust versions since they're so commonly used to test or support new compiler features.

eddyp: I was thinking of #[cfg] and associated macro https://doc.rust-lang.org/std/macro.cfg.html

ops/msrv.md
```diff
@@ -0,0 +1,49 @@
...
+
+- When doing a semver bump (i.e. minor version bump in pre-1.0 crates), for
+  whatever reason, the team in charge of the crate should consider bumping the
+  MSRV as well.
+
+- The MSRV will be documented in the crate level documentation of the crate,
+  like so:
+
+``` markdown
+# MSRV
```
Disasm: I suggest replacing `MSRV` with `Minimum supported Rust version` because `MSRV` abbreviation is unclear and Google doesn't know anything related about it.

thejpster:
I also agree with @eddyp 's comments, else it LGTM

ops/msrv.md
```diff
@@ -0,0 +1,49 @@
...
+
+- Changing the MSRV of a crate is a breaking change and requires a semver bump:
+  minor version bump if the crate is pre-1.0 and a major version bump if the
+  crate is (post-)1.0.
+
+- Cargo features are allowed to depend a on Rust version greater than the MSRV,
+  even a nightly compiler. For example, a "const-fn" can bump the required
+  version to 1.34.0 (effectively, nightly channel at the time of writing).
+
+- When doing a semver bump (i.e. minor version bump in pre-1.0 crates), for
```
Disasm: Are there any reasons for this? MSRV can be bumped with changes that need more recent Rust version and left as is the rest of the time.

japaric: I thought that it could help in a situation like this: say that the MSRV of your crate is currently 1.31, the latest stable release is 1.35 and that you are going to do a minor version bump. If you also bump the MSRV to 1.34 then you can, for example, use the Pin API (stabilized in 1.33) in the public API of the crate without having to bump the minor version (only a new patch release would be required) -- this is only helpful if you didn't plan / expect to use the Pin API when you did the minor version bump.

Disasm: Sounds reasonable!

ops/msrv.md
```diff
@@ -0,0 +1,49 @@
+# Minimum Supported Rust Version (MSRV)
+
+This text documents the MSRV policy used in the crates maintained by the WG.
+
+- The MSRV shall always be strictly smaller than the latest available minor
+  stable release. For example, if the latest stable Rust version was 1.32.1 a
+  new minor release of the `cortex-m` crate can't bump the MSRV higher than
+  1.31.0.
+
```
eddyp: It still bugs me a little, what if there is a 1.31.1? Typically, patch releases are supposed to be compatible 100% and are typically done to address issues with the x.y.0 release. So wouldn't it make sense to say 1.31.x?

japaric: You mean that the documentation should specifically say "MSRV is 1.31.**1**" / "MSRV is 1.31.**x**"? Newer patch releases can't break things so it doesn't make any difference whether the user uses 1.31.0 or 1.31.1. We might as well document this is as "MSRV is 1.31" -- the patch version doesn't matter.

eddyp: I think I mean the version example should be "1.31", so the MSRV should be specified as maj.min, instead of maj.min.patch.

eddyp: Hmm, can we use 1.31 instead of 1.31.0 in Travis config? Will it work as expected?

japaric: > Hmm, can we use 1.31 instead of 1.31.0 in Travis config?

No, the rust version must match rustup toolchain names, which use the `$major.$minor.$patch` syntax.

eddyp: So does anyone else thinks my comment makes sense? If so, how do we put in practice the verification on latest previous minor version? Is there any bot which could do the Travis config bump?

Disasm: I think it would be fine to use version x.y in MSRV text and x.y.0 in Travis config. If we state that crate is compatible with x.y, than it should be compatible with both x.y.0 and x.y.1.

ops/msrv.md
```diff
@@ -0,0 +1,49 @@
...
+This text documents the MSRV policy used in the crates maintained by the WG.
+
+- The MSRV shall always be strictly smaller than the latest available minor
+  stable release. For example, if the latest stable Rust version was 1.32.1 a
+  new minor release of the `cortex-m` crate can't bump the MSRV higher than
+  1.31.
+
+- Changing the MSRV of a crate is a breaking change and requires a semver bump:
+  minor version bump if the crate is pre-1.0 and a major version bump if the
+  crate is (post-)1.0.
```
posborne: One problem I have run into with several crates where I setup CI for a given MSRV over time is that even if nothing changes in the crate itself to break the MSRV when CI builds the project it uses the latest version of dependencies matching the Cargo.toml specifications as the project does not include one (as it is not an "application" itself).

Obviously, if every crate in the dependency chain followed this rule there is no problem (perhaps a good reason to keep deps to a minimum for central crates) but for a few of the Linux crates which have a few more dependencies I have run into this situation due to crates like nix violating this rule.  I guess the only solution in that case would be to use the version controls to enforce a `<` requirement.

Probably the only good solution to this problem is to get consensus throughout the rust community that an MSRV change is breaking.

japaric:
Current votes: 2 registered via GH (@Disasm @posborne) + 1 mine + 1 stale (@therealprof) + 2 in the comments (@korken89 @thejpster) = 7 / 27.

japaric:
Huzza! This RFC has received enough votes (9 registered + 1 mine of the required 33%) to be accepted.

Teams please start updating the crate level docs of your crates to indicate the MSRV.

bors +

japaric:
hmm

bors r+

Disasm:
Two notes:
1. MSRV policy does not work if at least one dependency doesn't obey it. Any patch version bump in the dependency tree may change overall actual MSRV.
2. Semver bump should be done not only for changing MSRV, but for defining MSRV too (for crates without defined MSRV)

Disasm:
It's not clear what should be done in the first case: the problem should be fixed somehow with patch version bump in order to preserve old MSRV.

PR #306, opened 2019-02-06
add team risc-v e-mail address

japaric:
r? @rust-embedded/riscv

### DIFF
--- a/README.md
+++ b/README.md
@@ -363,9 +363,9 @@ Each team can be privately contacted via the following e-mail addresses:
 - hal@teams.rust-embedded.org
 - infrastructure@teams.rust-embedded.org
 - resources@teams.rust-embedded.org
+- risc-v@teams.rust-embedded.org
 - tools@teams.rust-embedded.org
 <!-- - msp430@teams.rust-embedded.org -->
-<!-- - riscv@teams.rust-embedded.org -->
 
 You can usually find the members of the embedded WG on the #rust-embedded channel (server:
 irc.mozilla.org).


japaric:
By the way, I just realized that the e-mail address (risc-v@) and the team name (riscv) on GH don't match.

Disasm:
@japaric 
Yes, because team name is incorrect. See: https://riscv.org/risc-v-trademark-usage/

Disasm:
Also, I'm OK with `riscv` spelling in "technical" things like crate names, emails and URLs, but not with the human-readable names where `RISCV` spelling is used.

japaric:
bors=disasm

@Disasm should I rename the team then? There may be some breakage in .github/CODEOWNERS

Disasm:
@japaric if you can just change human-readable name from `RISCV` to `RISC-V`, it will be good. There should be no problems with CODEOWNERS if we leave `rust-embedded/riscv` internal name.

japaric:
bors r=disasm

@disasm AFAICT changing the human facing name to "RISC-V" changes the mention name to risc-v

Disasm:
@japaric Ok, then I suggest to leave this name as it is. It's not worth breaking the infrastructure :)

Disasm:
bors r+
Please!

japaric:
IDK what's wrong with bors it has been misbehaving in my personal repos.

The Travis build is [green](https://travis-ci.org/rust-embedded/wg/builds/489691353) (it's always green, actually ;-)), so I'm going to merge this manually.

Disasm:
Looks like I'm in RISC-V team, but not in All team, so "Permission denied" is not a bug. However, your `r?` should change the reviewer, but doesn't.

japaric:
@Disasm whoops, fixed.

"r?" does nothing because we are not using rust-highfive (yet).

PR #324, opened 2019-02-27
ops: hibernating

japaric:
This PR proposes a process that lets members 'hibernate' themselves when busy /
absent with the goals of not having highfive assign PRs to them and reducing the
number of votes required to get voting majority.

### DIFF
--- a/ops/hibernating.md
+++ b/ops/hibernating.md
@@ -1,0 +1,24 @@
+# Hibernating
+
+A member that will be absent or busy for an extended period of time and not able
+to participate in the WG during that time should put themselves in "hibernation"
+state. During this hibernation period:
+
+- The member will be removed from the highfive rotation. New PRs will *not* be
+  assigned to them by the highfive bot.
+
+- The member will be removed from their GitHub teams'. `cc @rust-embedded/$team`
+  will *not* cc them.
+
+- The member name will be removed from the list of teams in the rust-embedded/wg
+  README, but they will be listed under the 'Hibernating' section of that
+  README.
+
+- The member will not be counted when computing the number of votes required to
+  reach majority on matters that concern the teams they are a member of. This
+  includes PRs labeled `T-all` that need a decision.
+
+To enter or leave the hibernation state the member will:
+
+- Notify their teams (e.g. `cc @rust-embedded/$team`) and send a PR to
+  rust-embedded/wg updating the README to reflect their hibernation state.


ops/hibernating.md
```diff
@@ -0,0 +1,24 @@
+# Hibernating
+
+A member that will be absent or busy for an extended period of time and not able
+to participate in the WG during that time should put themselves in "hibernation"
+state. During this hibernation period:
```
Disasm: The problem is they will _not_ be able to put themselves in "hibernation" state if they are absent or busy. The only way to enforce this is to automate the process of hibernation or at least to send a reminder time to time.

hannobraun: It's possible to be able to (barely) keep up, while at the same time being too busy to really participate. This is my current situation, and if this PR is accepted before that situation changes, I'm going to use this mechanism to retire myself.

It's already possible to remove team members without their participation: https://github.com/rust-embedded/wg/blob/master/rfcs/0136-teams.md#removing-members

ops/hibernating.md
```diff
@@ -0,0 +1,24 @@
...
+
+- The member will be removed from the highfive rotation. New PRs will *not* be
+  assigned to them by the highfive bot.
+
+- The member will be removed from their GitHub teams'. `cc @rust-embedded/$team`
+  will *not* cc them.
```
hannobraun: Will the member stay part of `@rust-embedded/all`? Would be nice to be able to keep up with internal discussions while you're away (although those happen rarely anyway).

TeXitoi:
What is the problem this modification want to solve? Because hibernating looks like a heavy action.

For the problem of reaching a majority, the Google transit group have a different approach of voting: https://github.com/google/transit/blob/master/gtfs/CHANGES.md

TLDR:
 - Call for vote, everyone have one week to vote.
 - For vote to pass, you need at least 3 approvals and 0 rejection.
 - rejections must be motivated.

Like that, everyone have the time to approve, and if you don't give your opinion within 1 week, you don't block the system. The 3 minimal approvals allow to validate that the proposition is not controversial and useful.

Now, the Google transit WG is much more fuzzy: there is no list of members and anyone can vote.

ops/hibernating.md
Disasm: I think this can be achieved with tuning your notifications. We have "watch" button on each team page. So you can subscribe to /all and unsubscribe from everything else without any membership modifications.

ops/hibernating.md
japaric: > The problem is they will not be able to put themselves in "hibernation" state if they are absent or busy

Sure, but the proposal says that members that *will* be busy or absent (in the future) should put themselves into hibernation (now). This proposal is for them. And as @hannobraun said people currently busy may still be able to hibernate themselves.

> The only way to enforce this is to automate the process of hibernation

I don't think we should do this. Hibernation should be voluntary.

> at least to send a reminder time to time.

I don't think we should do this either. Sending an e-mail (once) to /all when / if this (or any other significant change) occurs seems fine, though.

ops/hibernating.md
japaric: @hannobraun no, /all is included in "their GitHub teams". We will not modify the member repository subscriptions, but IDK if removing someone from a team has any effects on their repository subscriptions.

japaric:
@TeXitoi 

> What is the problem this modification want to solve?

From the PR description:

"... with the goals of not having highfive assign PRs to them and reducing the number of votes required to get voting majority."

But more specifically the problem that this wants to solve is "you are busy and won't be able to participate for a while so you want to notify /all and be able to go absent without disrupting the operation of the EWG".

> For the problem of reaching a majority

We already have [a process] to reduce the 50% voting majority.

 FWIW, I think 1 week would be too fast for any kind of voting within the EWG. AFAIS most people tend to miss voting notifications and will only become aware of them at the weekly meetings.

[a process]: https://github.com/rust-embedded/wg/blob/master/rfcs/0206-voting-majority.md

jamesmunns:
Thank you all! This has now been accepted.

bors r+

PR #330, opened 2019-03-09
Update Resources Team Membership

jamesmunns:
I noticed that @rust-embedded/resources includes a larger list of people than the README. I can't remember if this was already decided, and the README just wasn't updated?

If this was already decided for @korken89 and @ryankurte, I don't think we need a full team vote. Could either of you please provide background?

### DIFF
--- a/README.md
+++ b/README.md
@@ -290,6 +290,8 @@ The resources team develops, maintains and curates resources on embedded Rust.
 - [@adamgreig]
 - [@andre-richter]
 - [@jamesmunns]
+- [@korken89]
+- [@ryankurte]
 - [@thejpster]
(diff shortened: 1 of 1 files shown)


jamesmunns:
If this was just a "provisional status before, and you're interested in joining full time, then I'm happy to vote "approve" here :)

korken89:
For now I'd like to be a bit unofficial, I have a bit too much on my table right now to provide help

ryankurte:
same tbqh, i'm also ok with being dropped if y'all would prefer not to have the inconsistency ^_^

jamesmunns:
Closing as this seems accurate.

PR #337, opened 2019-03-20
Remove linux-embedded-hal duplicat team assignment.

mathk:
Let me know if the linux-embedded-hal should be removed from the embbeded linux team instead of the HAL team.

### DIFF
--- a/README.md
+++ b/README.md
@@ -233,7 +233,6 @@ Board Support Crates and drivers.
 Projects maintained by the HAL team.
 
 - [`embedded-hal`]
-- [`linux-embedded-hal`]
 
 ### The MSP430 team
 


therealprof:
I just checked the repository and as it turns out the stated members here actually reflect reality, i.e. @rust-embedded/hal is a maintaining team for https://github.com/rust-embedded/linux-embedded-hal .

We probably should rectify that first before removing the from the wg page.

@rust-embedded/hal Opinions? Shall we remove ourselves from maintainership?
@rust-embedded/embedded-linux Would this be okay for you?

hannobraun:
> @rust-embedded/hal Opinions? Shall we remove ourselves from maintainership?

Sounds good to me. Having the HAL team maintain `linux-embedded-hal` doesn't make more sense to me than maintaining any other HAL implementation. Anyone from the HAL team who wants to maintain it, can just join the Embedded Linux team.

(then again, I'm not doing a whole lot of maintaining for anything these days, so feel free to disregard my opinion)

posborne:
Works for me.

therealprof:
@rust-embedded/core Can we treat this as an oversight and just correct it or do we want to RFC it?

jamesmunns:
@therealprof I'm happy with this just being a @rust-embedded/hal team vote.

- [x] @hannobraun 
- [ ] @ryankurte 
- [ ] @thejpster 
- [x] @therealprof 
- [ ] @ithinuel 

I'd say if we have one more vote, and no one objects, let's go ahead and just merge it. I don't think we need a full RFC here.

mathk:
Raise a PR to reflect this change : https://github.com/rust-embedded/linux-embedded-hal/pull/16

hannobraun:
We have a voting majority, and the other PR has been merged.

bors r+

Disasm:
bors r=hannobraun

PR #338, opened 2019-03-25
[RFC] `svd2rust` workspace generation

spacekookie:
As previously discussed at the Rust All-Hands 2019 in Berlin,
this RFC proposes to change the way that `svd2rust` generates platform crates,
away from peripheral modules to peripheral sub-crates.

[Rendered version](https://github.com/spacekookie/wg/blob/b3207b02d401b281c2d0f35d6cdff66133c2eb4a/rfcs/0000-svd2rust-workspaces.md)

### DIFF
--- a/rfcs/0000-svd2rust-workspaces.md
+++ b/rfcs/0000-svd2rust-workspaces.md
@@ -1,0 +1,122 @@
+# Svd2Rust workspace generation
+
+- Feature Name: svd2rust_ws_gen
+- Start Date: 2019-02-07
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+
+## Summary
+
+`svd2rust` is a tool to generate Rust code from svd-files.
+This allows easier development of embedded code for certain platforms and µ-controllers.
+There is a lot of code that is generated, sometimes upwards of a million lines of Rust.
+Currently `svd2rust` does this as a single crate, using modules to seperate different peripherals from each other.
+
+This RFC proposes a change to `svd2rust` to generate a workspace of crates, instead of just a single one.
+
+## Motivation
+
+Splitting up the code that is generated allows for ergonomics improvements in several dimensions.
+
+
+1. Users would be able to pick-and-choose crates they want to use, i.e. only supporting certain peripherals
+2. Using crates instead of modules allows `rustc` and `cargo` to use multiple codegen units,
+   making compiles parallel by default, using system resources more effectively.
+
+Especially decreased compile-times should be considered the primary motivation of this change.
+
+
+## Guide-level explanation
+
+Changes to `svd2rust` would mostly be under the hood. 
+As such, interaction with `svd2rust` after the change aren’t much different from how they are now. 
+
+Users can still just pull in a platform crate as a dependency → get all peripherals
+
+```toml
+    [project]
+    # ...
+    
+    [dependencies]
+    acme_mcu8008 = "1.1.0" # Includes _all_ peripherals
+```
+
+Additionally it’s possible to depend on a specific sub-set of peripherals
+
+  - Pull in common platform crate
+  - Only get access to specific peripherals
+
+```toml
+    [project]
+    # ...
+    
+    [dependencies]
+    acme_mcu8008_ethernet = "1.1.0"
+    acme_mcu8008_spi = "1.1.0"
+    acme_mcu8008_wom = "1.1.0"
+```
+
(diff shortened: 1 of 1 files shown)


jamesmunns:
CC (especially) @therealprof, @Emilgardis, @japaric, @rust-embedded/tools.

rfcs/0000-svd2rust-workspaces.md
```diff
@@ -0,0 +1,122 @@
...
+    [dependencies]
+    acme_mcu8008_ethernet = "1.1.0"
+    acme_mcu8008_spi = "1.1.0"
+    acme_mcu8008_wom = "1.1.0"
+```
+
+The usage of modules `ethernet`, `spi` and `wom` would then only differ by their top-level import name.
+
+```rust
+    use acme_mcu8008::{ethernet, spi, wom};
```
japaric: Presumably today's `Peripherals` singleton struct which contains all peripherals will live in this crate. If one does not want to use that crate to get faster compile times how do they get an instance of say `USART1`? Will `USART1` gain a `take` static constructor whose implementation contains a  `static ONCE: bool` to prevent getting several instances of the same peripheral / singleton? That would use more RAM than today's implementation and should be noted.

rfcs/0000-svd2rust-workspaces.md
```diff
@@ -0,0 +1,122 @@
...
+- When using `codegen-units=1` in release mode, this could have impact on compile-times
+- Possibly lower optimisation potential because of using multiple codegen units
+
+Especially the last two points would require some proper profiling rather than speculation
+to come to a conclusion about how much performance impact (and in what direction) this feature will have.
+
+
+## Rationale and alternatives
+
+An alternative to generating multiple crates for peripherals was to generate feature flags for each peripheral,
```
japaric: Perhaps mention here that both approaches lead to same amount of code being compiled but this proposal has the advantage of parallel compilation of peripheral code. It would be interesting to have numbers on how much parallel compilation helps with overall compile times.

rfcs/0000-svd2rust-workspaces.md
```diff
@@ -0,0 +1,122 @@
...
+[PR #236](https://github.com/rust-embedded/svd2rust/pull/236) and [PR #254](https://github.com/rust-embedded/svd2rust/pull/254). 
+An unpublished version of `svd2rust` workspace generation exists and highlighted a few issues that have
+influenced the design of this RFC.
+A tool to handle workspace-synchronised releases has a partially working implementation
+[here](https://github.com/spacekookie/cargo-ws-release).
+
+
+## Unresolved questions
+
+A HAL crate could potentially have dozens of peripheral sub-crates.
```
japaric: For the record I have been contacted by the docs.rs team before because svd2rust crates were clogging the build queue so they ended up disabling doc generation for targets other than Linux (normally, `cargo doc` is run on your crate several times to generate docs for Linux, Windows, macOS, etc.) on these crates to reduce the build times.

I wonder if sequentially building docs for 50 crates w/ 20K LoC each takes longer than building the docs of a single crate w/ 1M LoC.

adamgreig:
At [stm32-rs](https://github.com/stm32-rs/stm32-rs/) we build svd2rust crates for all STM32 devices, factoring out the common changes to svd files to help reduce duplication of effort. The svd2rust output is turned into a single module per device, and then all devices belonging to the same family are published as part of the family crate. For example, the stm32f405 device is a module in the stm32f4 crate, which is on crates.io. This means there's 11 crates which is already a fair amount to manage -- even with a script publishing them sequentially, it takes some time to publish and then docs.rs takes a real hit each release. In fact we only generate documentation for one or two devices per family to avoid overloading docs.rs.

Overall we have 48 devices so far, and in total over 1500 peripherals, so with this change we'd go from 11 crates to >1500, just for this project. I think it would take it beyond manageable even with automation -- that would be over 6% of all of crates on crates.io!

I _totally_ get wanting to reduce compile times, but for the family projects like stm32-rs I don't think a crate per peripheral is the way to go, if anything I'd like to see built-in support for generating _fewer_ crates. Ideally I'd just have a single stm32-rs crate with every stm32 supported and all the common peripherals automatically factored out!

Perhaps if this is added it could be an optional switch, so at least stm32-rs could keep doing what it does? I like feature-gating the peripherals as a way to reduce compilation times without parallelism and it seems like feature gates would also integrate well with HAL crates providing feature-gated implementations per peripheral. The additive nature of features works perfectly with this approach too.

PR #350, opened 2019-05-17
Add "Hibernating" section to README, add myself

hannobraun:
I haven't been very active in the working group for a long while now, and I don't see this changing much going forward. For this reason, I'm putting myself into the hibernation state.

I'm not leaving the community though! I'm still involved in [nrf-rs](https://github.com/nrf-rs), [lpc-rs](https://github.com/lpc-rs), and my various other open source projects. I also intend to still keep up with many of the WG repositories, so the main difference is going to be that I'll no longer feel bad for not doing any actual work :-)

cc @rust-embedded/hal

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,7 +222,6 @@ Board Support Crates and drivers.
 
 #### Members
 
-- [@hannobraun]
 - [@thejpster]
 - [@therealprof]
 - [@ithinuel]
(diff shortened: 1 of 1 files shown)


hannobraun:
I was waiting for someone to remove me from the teams, but I just realized I can do this myself.

bors r=ryankurte (this might not work anymore, not sure when/how permissions are updated)

ryankurte:
hmmm

bors r+

PR #363, opened 2019-06-26
Updating Ops from IRC to Matrix

adamgreig:
# Updating Ops from IRC to Matrix

This PR replaces the existing IRC ops notes with a new Matrix ops note. I've
also put notes about migration details here for discussion. I'll aim to update this top post with any changes due to discussion here, before we merge.

## Summary

We are migrating from our current IRC channel on Mozilla IRC, which is shutting
down in the near future, to a new Matrix room. We'll migrate as soon as
practical after a trial period. We'll register a new IRC channel on Freenode
which we bridge to the Matrix room.

## Proposed Timeline

* Set up the Matrix room and IRC bridge as described in this PR as soon as
  possible
* Begin a two week trial period once the room is set up
* Collect feedback at the end of the two week trial from all users and address
  any outstanding concerns or issues. Make a decision at this point on whether
  any services need to be self-hosted, e.g., a homeserver, an IRC bridge,
  or public logging.
* As soon as possible after all issues are resolved, officially migrate from
  IRC to Matrix. Advertise migration.
* Update the Mozilla IRC channel topic to point to Matrix, and set the
  channel +m (moderated: prevents non-ops from talking).

### Room Setup

* Room name, topic, avatar, and settings are all already configured
* Administrators and moderators need adding
* IRC (and Discord?) bridges need setting up
* Some sort of public web-facing logging needs setting up

### Advertisement

We need to update the following resources:

* Documentation which points to the IRC channel (done in this PR)
* Weekly agenda Dropbox Paper
* Code of Conduct files on other WG repositories (done for wg repo in this PR)
* [README on Awesome Embedded Rust](https://github.com/rust-embedded/awesome-embedded-rust/blob/master/README.md)

We should advertise the change in the following places:

* Our Twitter channel
* In a blog post
* In the old IRC channel topic
* The [rust-embedded homepage](https://github.com/rust-embedded/rust-embedded-www) should be updated to add a link

### Freenode Registration

We would register as the Rust Embedded group on Freenode
([groupreg](https://freenode.net/groupreg).

A member of the core team would contact Freenode's Group Management & Community
Team to apply for registration and the `#rust-embedded` channel. The other core
team members should be added as group contacts. Channel operators would then be
added as required.

### IRC Bridging

IRC was the second-place option in our vote, and is our current communication
method, so we would like to continue to provide an IRC option for users while
migrating to Matrix as the official room. Matrix provides IRC bridging, where
users in both bridged rooms can interact as though they all shared a single
room.

We would bridge to the `#rust-embedded` channel on Freenode, pending its
registration as discussed above. Once registered, a core team member (both an
admin in the Matrix room and an operator on the IRC channel) can add the bridge
from their Matrix client. No special settings are required, and we would use
the public matrix.org-hosted IRC bridge.

We plan to maintain the IRC bridge to Matrix unless it becomes an
administrative burden, for example due to abuse or spam.

### Self-Hosting

During the 2019-06-25 meeting there was some discussion to self-hosting a
Matrix homeserver. Homeservers allow users to register and access the Matrix
federated network through that server. Matrix rooms themselves are federated
throughout the network and do not live in one specific server, so we would not
be hosting our own room per se, but allowing WG or community members to
register with our home server might provide better performance for those users.

All users could continue to access the room from their own home server, which
for many users would likely be the public matrix.org homeserver.

This PR proposes we do not initially self host due to the overhead of running
our own server. During the trial period we might discover that performance is
not suitable using the public homeservers and investigate self-hosting.

We could also consider self-hosting an IRC bridge. This PR proposes using the
public matrix.org IRC bridge initially.

### Web-based logging

We would like to set up a web-accessible room log; there already exists
suitable software
([matrix-static](https://github.com/matrix-org/matrix-static))
but is not clear if we'll have to host this ourselves or can use a public
hosted instance.

### Discord Bridge

Discord was the third-place option in our vote, so it clearly has some support
in the community.

There is an option to set up a bridge to a Rust Embedded Discord channel as
well, but this has not yet been investigated thoroughly. We would require
permission from the Discord administrators.

## Alternatives

* Don't bother with bridges. It may turn out that everyone is happy enough
    using Matrix directly, or that the time or effort cost of running the
    bridge is prohibitive.
* Bridge to Mozilla IRC while it continues to operate, then update bridge to
    Freenode afterwards. I think doing this will just add more administration
    and we'd be better off making a clean split with timing under our control,
    but it's likely the Mozilla IRC will remain online for some time so it
    might be preferable to bridge to our current room at first.
* A new team for managing IRC/Matrix, rather than using the core team.


ops/matrix.md
```diff
@@ -0,0 +1,77 @@
...
+* Can send messages
+* Can invite users
+
+## IRC Bridge
+
+The Matrix room is bridged to an IRC channel, allowing IRC users to communicate
+with users in the Matrix room. This is provided as a convenience to IRC users
+and the WG may revisit the bridge in case it compromises the Matrix users experience.
+
+Our IRC channel is `#rust-embedded` on [Freenode](https://freenode.net).
```
x37v: shouldn't this be `##rust-embedded` ??

x37v: ahh, once it becomes official it'll be `#rust-embedded` ?

jonas-schievink: It's official now :tada:

adamgreig:
Hi all, I'd like to push this towards merging!

Current status is:
* Matrix room is operational and has been in use for something like 2 months, with ~164 users
* IRC channel #rust-embedded on Freenode is registered to us and bridged to Matrix
* Public logging - via the IRC channel - is running on logbot as before (https://freenode.logbot.info/rust-embedded)
* The old IRC room is now unused and points to the new Matrix and IRC rooms
* Nothing is self hosted

I think all we really need to do is update links on our repos and a quick advertising blitz on Twitter and the blog.

Does anyone have any comments or can we move towards merging this PR?

thejpster:
I second Adam's comments above.

Disasm:
@adamgreig do you mind updating `(TBC)` logging link?

adamgreig:
@Disasm good point, updated.

adamgreig:
Ok, I think that's enough approval!

bors r+

PR #376, opened 2019-09-05
add @almindor to the RISC-V team

almindor:


burrbull:
@Disasm what your thoughts?

PR #377, opened 2019-09-10
[RFC] Add a Mutex trait

korken89:
cc @perlindgren @japaric 

[Rendered](https://github.com/korken89/wg/blob/master/rfcs/0377-mutex-trait.md)


roblabla:
I prefer a closure. There are certain kinds of locks that are not implementable by a guard because the destructor can be moved around and reordered. For instance, locks that backup and disable the current status of interrupts cannot (easily) be reordered.

It's possible to implement a closure-based lock on top of a guard lock, but not the other way around. E.G.

```
fn lock(&self, closure: F) {
    let guard = self.inner_lock.lock();
    closure(&mut *guard);
}
```

As such, I think a closure-based design is better.

andre-richter:
>> because the destructor can be moved around and reordered.

You mean the compiler moving around the destructor?

korken89:
From my side there are 4 reasons why this should not be based on RAII, but a closure:

1. @roblabla already touched on, the RAII version can be used to implement the closure and hence the closure can be implemented with an `std`-like mutex in the background.
2. The closure is explicit where it starts and ends, no need to wait for the destruction or add extra `{ ... }` around code to get predictable unlocks. 
3. Makes code using locks more stable under refactoring, as small changes can drastically change the scope of a lock when using RAII.
4. A RAII based mutex can be `mem::forget` and all guarantees are out, the closure based mutex cannot be circumvented easily.

Lets continue discussing, I'll update the RFC with these thoughts after. :)

roblabla:
@andre-richter I mean you can move around the Guard.

E.G.

```rust
let x = lock1.lock();
let y = lock2.lock();
drop(x);
drop(y);
```

Here, lock1 is locked after lock2, but lock1 is unlocked first. This might work for some types of mutexes, but others will be hard to implement this way.

andre-richter:
Yeah, like I said, I'm in favor of closures as well. Thanks for providing the feedback so far. I just wanted to point out that users coming from standard world might be confused for a second but it's a price we can easily pay.

In the past I wanted to check/benchmark if RAII vs. closure has any (minimal) performance differences.

If somebody feels like writing a micro benchmark, this would be a nice opportunity now and potential content for the disadvantages section.

korken89:
After inlining the performance should be the same as they fundamentally do the same operations, but if someone wants to benchmark I can add the results.
We know from experiments in RTFM that the closures inline and disappear at least, but I have no data for RAII mutexes.

Another thing that came up with RAII based Mutex, if someone runs `mem::swap` on two mutexes - what does this mean?

andre-richter:
Yeah I was actually wondering if there are situations where the compiler would decide to call the closure as a function.

Agree that with inlining it is zero overhead.

We could put a note about what is needed (if any) to guarantee that inlining is done, or what should be avoided in a release build b/c it would defeat inlining.

roblabla:
I've found https://github.com/japaric/cortex-m-rtfm/blob/2596ea0e46bec73d090d9e51d41e6c2f481b8e15/book/en/src/internals/critical-sections.md which explains in much finer details the reasoning behind `&mut T` for the mutex (with some clear examples of what it enables). Maybe this should get mentioned in the RFC.

korken89:
Thanks, I will add it for clarification

rfcs/0377-mutex-trait.md
```diff
@@ -0,0 +1,243 @@
...
+[drawbacks]: #drawbacks
+
+As of writing this RFC I see no drawbacks.
+
+# Alternatives
+[alternatives]: #alternatives
+
+The alternatives, taking `&self` and returning a `Result<...>` are discussed in [Detailed design](#detailed-design).
+
+One can also futher follow the [Rust standard library `Mutex`] and its RAII pattern, however this should not be persued dues to these causes:
```
andre-richter:
```suggestion
One can also further follow the [Rust standard library `Mutex`] and its RAII pattern, however this should not be pursued dues to these causes:
```

rfcs/0377-mutex-trait.md
```diff
@@ -0,0 +1,243 @@
...
+
+1. The RAII version can be used to implement the closure and hence the closure can be implemented with an `std`-like mutex in the background.
+2. The closure is explicit where it starts and ends, no need to wait for the destruction or add extra `{ ... }` around code to get predictable unlocks.
+3. Makes code using locks more stable under refactoring, as small changes can drastically change the scope of a lock when using RAII.
+4. A RAII based mutex can be `mem::forget` or `mem::swap` and all guarantees are out, the closure based mutex cannot be circumvented easily.
+5. The RAII based mutex can be circumvented easily as follows:
```
andre-richter:
```suggestion
  4.1. The RAII based mutex can be circumvented easily as follows:
```

rfcs/0377-mutex-trait.md
```diff
@@ -0,0 +1,242 @@
...
+[drawbacks]: #drawbacks
+
+As of writing this RFC I see no drawbacks.
+
+# Alternatives
+[alternatives]: #alternatives
+
+The alternatives, taking `&self` and returning a `Result<...>` are discussed in [Detailed design](#detailed-design).
+
+One can also further follow the [Rust standard library `Mutex`] and its RAII pattern, however this should not be persued due to these causes:
```
andre-richter:
```suggestion
One can also further follow the [Rust standard library `Mutex`] and its RAII pattern, however this should not be pursued due to these causes:
```

andre-richter:
Follow up points from discussion on Matrix:

- Trait bounds for indicating properties like single/multi-core usage, etc.

therealprof:
It think it should be an `unsafe trait` due to the extra care required by the implementer to implement it soundly, as mentioned in https://github.com/rust-lang/rfcs/blob/master/text/0019-opt-in-builtin-traits.md:

> Unsafe traits: An unsafe trait is a trait that is unsafe to implement, because it represents some kind of trusted assertion. Note that unsafe traits are perfectly safe to use. Send and Share are examples of unsafe traits: implementing these traits is effectively an assertion that your type is safe for threading.

There's also precedence in the `core` lib about that (https://doc.rust-lang.org/src/core/str/pattern.rs.html#94) and a helpful comment outlining the reasoning:

> /// The trait is marked unsafe because the indices returned by the
> /// `next()` methods are required to lie on valid utf8 boundaries in
> /// the haystack. This enables consumers of this trait to
> /// slice the haystack without additional runtime checks.

roblabla:
> There's also precedence in the core lib about that (https://doc.rust-lang.org/src/core/str/pattern.rs.html#94) and a helpful comment outlining the reasoning:

There's a difference though. This trait is unsafe because other code is allowed to assume the returned value is on a valid utf8 boundary. If an implementation fails to respect this, UB and memory corruption are likely to occur. This requirement is not actually encoded in the type system, so `unsafe` is necessary.

In our case, the requirement is that the closure is "guaranteed exclusive access to the data contained within the mutex for the duration of the lock." But this is already guaranteed by the type system: We give the closure an `&mut T`, which is by definition an exclusive access to the data contained.

perlindgren:
@roblabla , I'm not sure I understand. To implement the Mutex trait, should be unsafe (we have to take responsibility that the `lock` function gives exclusive access for the duration of the lock (in our case the closure). The clousure itself does not need to be unsafe to use, just to implement trait is unsafe.

/Per

roblabla:
@perlindgren All of the requirements for safe usage of this trait are encoded by the type system. Safe usage of this trait requires that the closure has exclusive access to T. The closure gets a &mut T. An `&mut T` is an exclusive reference. It is impossible in safe code to create two live `&mut T` to the same T. The only way to violate this guarantee is through broken unsafe code.

If `Mutex` had other invariants that aren't encoded by the type-system, then it would need to be an `unsafe trait` as a purely safe implementation might not catch those invariants. But as it stands, Rust already has all the compile-time checks to ensure that Safe Rust cannot implement this trait incorrectly, and as such it should not be unsafe.

RTFM has an example of a *completely safe* implementation of Mutex called Exclusive: https://github.com/japaric/cortex-m-rtfm/blob/master/src/lib.rs#L311. It's trivial to know this implementation of Mutex is correct *because* the invariants that the trait must guarantee are encoded in the type system.

perlindgren:
@roblabla , yes you are completely correct, the use of the Mutex lock is completely safe and checked by the type system. RTFM guarantees deadlock freedom, building on earlier work on the Stack Resource Policy (Baker 90) and our C code implementation (2013 and on...).

The thing here is not the use of the Mutex lock, but the actual implementation of the trait. You should not be allowed to implement the trait in safe code (as when you do, you take the responsibility for its correctness which is out of reach for the Rust type system to verify).

roblabla:
I fail to see which guarantee is out of reach for the type system. If there are any, they are not currently present in the RFC and need to be documented.

korken89:
@roblabla Text and comments added on `unsafe`

rfcs/0377-mutex-trait.md
```diff
@@ -0,0 +1,261 @@
...
+
+It boils down to this philosophy: `panic` is for program error - `Result` is for user error
+
+#### Why is the trait `unsafe`?
+
+The trait is marked `unsafe` in accordance with [RFC 19, 127], which states:
+
+> Unsafe traits: An unsafe trait is a trait that is unsafe to implement, because it represents some kind of trusted assertion. Note that unsafe traits are perfectly safe to use. Send and Share are examples of unsafe traits: implementing these traits is effectively an assertion that your type is safe for threading.
+
+In this case the trusted assertion is that the data protect by the mutex is only accessed in a critical section, as we are going to share mutable references to the internal data. If this is not upheld by the implementation i.e. an unsound mutex implementation, the memory will be aliased mutably which is undefined behaviour. As the core requirement cannot be expressed using the Rust type-system in many cases, this requirement is placed on the implementor to uphold.
```
therealprof:
```suggestion
In this case the trusted assertion is that the data protected by the mutex is only accessed in a critical section, as we are going to share mutable references to the internal data. If this is not upheld by the implementation i.e. an unsound mutex implementation, the memory will be aliased mutably which is undefined behaviour. As the core requirement cannot be expressed using the Rust type-system in many cases, this requirement is placed on the implementor to uphold.
```

andre-richter:
Just thinking out loud:
Should we provide marker traits that an implemetor can use to signal if a mutex is safe in multi-core environments (as opposed to only being usable in single-core env)?

roblabla:
I still do not understand why this trait needs to be unsafe. This trait makes no guarantees that are not checked by the type system. If it does, then the [current Exclusive implementation](https://github.com/japaric/cortex-m-rtfm/blob/master/src/lib.rs#L311) in RTFM is unsound. And I'd like it if someone could point where the unsoundness in RTFM's Exclusive impl is.

> If this is not upheld by the implementation i.e. an unsound mutex implementation, the memory will be aliased mutably which is undefined behaviour.

Please show how this can happen in Safe Rust. The whole point of `&mut T` is that they cannot alias mutably. If they do, then something is already unsound before Mutex is involved. It is illegal in rust to have aliasing mutable borrows.

--------

The only new guarantee being made here is :

> In this case the trusted assertion is that the data protected by the mutex is only accessed in a critical section. [snip]  As the core requirement cannot be expressed using the Rust type-system in many cases, this requirement is placed on the implementor to uphold.

Which is different (and stronger) from what is currently being asserted in the Rules for Implementation section, which only cares about exclusive access *for the duration of the lock*. Here the core requirement is that the implementor must **not** modify the data underlying the mutex on its own without going through a "Critical section". This seems like a silly requirement, since the mutex implementor is in charge of defining the critical section anyways.

TeXitoi:
I also don't think that this trait must be unsafe.

rfcs/0377-mutex-trait.md
```diff
@@ -0,0 +1,261 @@
...
+fn unsound(mtx: &impl core_mutex::Mutex<Data = i32>) {
+    mtx.lock(|data| {
+        *data += 1;
+
+        mtx.lock(|data| *data += 1);
+    });
+}
+```
+
+This example allows for a mutex to lock itself within a lock of itself, which is **unsound by construction** and creates an immediate deadlock - while if implementing it with `&mut self` the following would happen:
```
TeXitoi: That's not unsound, it should deadlock, which is sound.

rfcs/0377-mutex-trait.md
```diff
@@ -0,0 +1,261 @@
...
+
+#[interrupt]
+fn SWI0_EGU0() {
+    let mut r = &MUT; // Note that r is mutable
+
+    // Locking!!
+    r.lock(|data| *data += 1);
+
+    // More Locking!!
+    increment_in_mutex(&mut r);
```
TeXitoi: `&mut` not needed here.

TeXitoi: Sorry bad reading

rfcs/0377-mutex-trait.md
```diff
@@ -0,0 +1,261 @@
...
+
+#### Why is the trait `unsafe`?
+
+The trait is marked `unsafe` in accordance with [RFC 19, 127], which states:
+
+> Unsafe traits: An unsafe trait is a trait that is unsafe to implement, because it represents some kind of trusted assertion. Note that unsafe traits are perfectly safe to use. Send and Share are examples of unsafe traits: implementing these traits is effectively an assertion that your type is safe for threading.
+
+In this case the trusted assertion is that the data protected by the mutex is only accessed in a critical section, as we are going to share mutable references to the internal data. If this is not upheld by the implementation i.e. an unsound mutex implementation, the memory will be aliased mutably which is undefined behaviour. As the core requirement cannot be expressed using the Rust type-system in many cases, this requirement is placed on the implementor to uphold.
+
+Another way to view this is that an unsound implementation of a mutex can lead to aliasing as the assumption made on the guard having `&mut self` is not as strict as one would expect from the Rust type-system.
```
TeXitoi: I don't think it should be an unsafe trait as having this aliasing is unsound by definition, and thus no need do implement it unsafely.

rfcs/0377-mutex-trait.md
```diff
@@ -0,0 +1,261 @@
...
+# Alternatives
+[alternatives]: #alternatives
+
+The alternatives, taking `&self` and returning a `Result<...>` are discussed in [Detailed design](#detailed-design).
+
+One can also further follow the [Rust standard library `Mutex`] and its RAII pattern, however this should not be pursued due to these causes:
+
+1. The closure is explicit where it starts and ends, no need to wait for the destruction or add extra `{ ... }` around code to get predictable unlocks.
+2. Makes code using locks more stable under refactoring, as small changes can drastically change the scope of a lock when using RAII.
+3. A RAII based mutex can be `mem::forget` or `mem::swap` and all guarantees are out, the closure based mutex cannot be circumvented.
```
TeXitoi: guarantees are not out, it will just not unlock, and thus deadlock if another lock is done, which is sound.

rfcs/0377-mutex-trait.md
```diff
@@ -0,0 +1,261 @@
...
+    * A RAII based mutex can be circumvented easily as follows:
+
+```rust
+// Break all guarantees
+let x = lock1.lock();
+let y = lock2.lock();
+drop(x);
+drop(y);
```
TeXitoi: I don't understand why it break all guarantees.

roblabla: Imagine a lock that works in a "backup-and-restore" way for interrupts.

1. The first lock would backup the current Interrupt Flag (Active) and set it to Inactive.
2. The second lock would backup the current Interrupt Flag (Inactive) and set it to Inactive.
3. Unlocking the first lock would restore its saved Interrupt Flag (Active)
4. Unlocking the second lock would restore its saved Interrupt Flag (Inactive)

Of course it's possible to avoid this by implementing things differently (such as keeping a global depth count or something), but the closure-based design makes it much easier to avoid shooting yourself in the foot.

TeXitoi: Then, that's more than RAII complexify the implementation of some mutex, or even avoid some implementation.

andre-richter:
Please look up actual mutex implementations, eg in the spin crate.
Aliasing mutable borrows IS how it usually is done. 
You can't implement what we need without unsafe.

roblabla:
@andre-richter this is irrelevant. Just because unsafe is necessary to **implement** doesn't need the trait itself is unsafe. An unsafe trait means the trait enforces guarantees that are **outside** the type system. This is not the case here. Just because the user is likely to need unsafe to implement the trait does not mean the trait should be marked unsafe.

andre-richter:
Would a guarantee outside the type system be that the code can assure that the Hardware is in a certain state,  for example IRQs are disabled?

adamgreig:
I also don't think this needs to be an `unsafe` trait. The point of marking a trait `unsafe` is to ensure you cannot implement it without `unsafe` code, to prevent the possibility of UB from only safe code. For `Send` and `Sync`, that's required because the compiler assigns those marker traits special meanings around concurrent access. For our `Mutex`, there's no special meaning assigned beyond the normal use of mutable references, as @roblabla pointed out, so you cannot cause UB just by implementing this mutex incorrectly in safe code.

Or to put it another way: if an entirely safe implementation of our Mutex trait is able to deliver a `&mut` to the closure, without using any `unsafe` code, then the whole of Rust's type system has already guaranteed that the reference is unique, so the Mutex obligations are met. We should only require that our trait is `unsafe` if an otherwise safe implementation causes UB, which I don't believe is possible with this design. I'm happy to see an example otherwise though!

roblabla:
> Would a guarantee outside the type system be that the code can assure that the Hardware is in a certain state, for example IRQs are disabled?

Yes. It'd make the trait kind of platform-specific though ^^. And under such a model, RTFM's Exclusive would not be a valid Mutex implementation.

adamgreig:
> Would a guarantee outside the type system be that the code can assure that the Hardware is in a certain state, for example IRQs are disabled?

Even _with_ such a requirement in our Mutex trait design, I still don't think we should mark it as `unsafe` -- requiring IRQs are disabled has no bearing on whether the trait is safe to implement. We'd only need it if users of our Mutex were able to do unsafe things in safe code using the Mutex.

Even if you assume that inside the closure all interrupts are disabled, but in fact they were not, what UB could you trigger using only safe code?

roblabla:
> requiring IRQs are disabled has no bearing on whether the trait is safe to implement.

I disagree. If such a requirement is placed on mutex, then unsafe code in the closure would be allowed to assume that it cannot be interrupted and do many fancy things that would completely break down and potentially cause UB if it actually got interrupted.

adamgreig:
> > requiring IRQs are disabled has no bearing on whether the trait is safe to implement.
> 
> I disagree. If such a requirement is placed on mutex, then unsafe code in the closure would be allowed to assume that it cannot be interrupted and do many fancy things that would completely break down and potentially cause UB if it actually got interrupted.

But I don't think that's what `unsafe` should mark on a trait. Once the closure contains unsafe code anyway it's outside of Rust's safety system. It's definitely an _error_ for the mutex implementation to not really disable interrupts, but it doesn't have a bearing on the _safety_, since you still need unsafe code in the closure to trigger any UB.

roblabla:
> Once the closure contains unsafe code anyway it's outside of Rust's safety system

Unsafe code just says the author manually checked all the invariants, and they are upheld. If the closure's unsafe code relies on the lack of IRQ to not be unsound, then *something* needs to assert it. This could simply be an assert before the offending unsafe code, or this can be an invariant given by an unsafe trait. 

Say you have a trait with an invariant "must disable IRQ". If the trait is marked safe, it's totally admissible to make an implementation that simply doesn't implement the invariant. It should be assumed that every safe rust implementation of a safe trait are allowed. As such, you can't write Unsafe Code that relies on the safe trait invariants living outside the type system to be upheld.

adamgreig:
> Say you have a trait with an invariant "must disable IRQ". If the trait is marked safe, it's totally admissible to make an implementation that simply doesn't implement the invariant.

I don't see why this is any less a bug than if the trait was marked unsafe. Either way, an error in the trait implementation leads to UB in the unsafe user code, but whether the trait is marked unsafe or not has no bearing on whether the user code is able to cause UB. My understanding is the trait needs to be marked unsafe if it allows users to cause UB in safe code, which wouldn't be the case in the IRQ-disabling example. What additional invariants must be upheld is part of the design contract for the trait, but not part of the safety system.

Anyway though, I don't think we actually want to add such a restriction here, right? It would be useful and possible to implement an embedded Mutex using atomics with no disabling interrupts at all, and we already have `bare_metal::CriticalSection` for such a thing.

roblabla:
> My understanding is the trait needs to be marked unsafe if it allows users to cause UB in safe code, which wouldn't be the case in the IRQ-disabling example.

There's a synergy between unsafe trait and unsafe code. Safe code is irrelevant in the whole thing - `unsafe trait` can't really be used by safe code anyways, since they don't have the ability to do potentially unsafe things with it. To take the words from [RFC19](https://github.com/rust-lang/rfcs/blob/master/text/0019-opt-in-builtin-traits.md):

> An unsafe trait is a trait that is unsafe to implement, because it represents some kind of trusted assertion

This assertion is *trusted* by unsafe code to be upheld. Let's take the [Searcher](https://doc.rust-lang.org/core/str/pattern/trait.Searcher.html) example. Searcher guarantees that "the indices returned by the next() methods [...] lie on valid utf8 boundaries in the haystack." It further notes that "This enables consumers of this trait to slice the haystack without additional runtime checks."

In the libcore, we have unsafe code in the [trim_matches function](https://doc.rust-lang.org/src/core/str/mod.rs.html#3705-3708) that relies on the guarantees asserted by the Searcher trait to be correct. Specifically, it uses the unsafe `get_unchecked` function with the indices returned by the `Searcher::next` function. If the searcher returns indices outside utf-8 sequence boundaries, then this would be UB.

Yet we consider this piece of unsafe code to be correct. This is because `Searcher` is an unsafe trait, and its implementations **must** uphold the guarantees laid out in the contract. This doesn't mean the `Searcher` trait must be implemented with unsafe code. It merely means it **must** adhere to additional contracts beyond what the type system provides.

If searcher was a safe trait, this piece of code would be considered unsound and would need to be rewritten. But because it is an unsafe trait, if an implementation fails to uphold the contract, then it is that implementation that would be considered unsound. Seen this way, it's a blame assignment issue.

> Anyway though, I don't think we actually want to add such a restriction here, right?

I agree with you here. Mutex shouldn't have such a restriction.

perlindgren:
I would still argue the trait should be unsafe. This is the only way we can alarm the user that the trait implies the extra guarantees that is implied by the Mutex abstraction. Think of the Sync trait. Its unsafe, since it implies guarantees/side effects that cannot by guaranteed by the Rust type system. Although the trait itself is just a marker (from your perspective it should be a safe trait right, .... but its not). 

Rust as of today provides us with just "one line of defense" and that is the safe/vs unsafe. One could perhaps think of inventing some other keyword, perhaps "system" for system level I/O, and other things that should never be written in safe user code. 

The Mutex trait handling out mutable references is however unsafe, as in our setting we have to expect that multiple instances of mutable resource "hooks" are existing coexists. 

Regarding multiple locking, (with the effect of deadlock), is indeed from a type system perspective sound, BUT highly undesirable. By making the Mutex take an &mut self, solves this problem in an elegant manner, and leverages the Rust type system to reject such cases at compile time. (RTFM Mutex gives an even stronger guarantee of total deadlock freedom at zero cost). Stepping back from such guarantees would be devastating, and throw as back to the times of thread based programming.

roblabla:
@perlindgren 

> This is the only way we can alarm the user that the trait implies the extra guarantees that is implied by the Mutex abstraction.

Please elaborate which guarantees. As I have already exposed above, there are none.

> Think of the Sync trait. Its unsafe, since it implies guarantees/side effects that cannot by guaranteed by the Rust type system

No, the Sync trait should absolutely not be safe. It is an unsafe trait because unsafe code is allowed to rely on Sync objects being allowed to be shared between threads. This is an additional guarantee that would not exist in the type system if Sync wasn't there. If sync was a safe trait, anyone would be allowed to add `Sync` to their types without having to sync, and all unsafe code relying on the invariants imposed by Sync would be unsound.

adamgreig:
> Think of the Sync trait. Its unsafe, since it implies guarantees/side effects that cannot by guaranteed by the Rust type system. Although the trait itself is just a marker (from your perspective it should be a safe trait right, .... but its not).

The Sync trait has to be unsafe because if it were safe to implement, then safe code could easily cause UB. Implementing it is telling the compiler that your implementation maintains the required invariants, without which safe user code could cause UB, because the compiler treats Sync specially (e.g., all `static` items must be `Sync`). Our Mutex trait isn't a special case to the compiler.

> The Mutex trait handling out mutable references is however unsafe

I'm not sure we're clear on the distinction between implementors of the trait (who are the only ones concerned with whether the trait is unsafe) and users of the trait, who wouldn't have to write unsafe either way. An implementation that hands out mutable references might have to use unsafe code under the hood, but by the time it's presented the user with an `&mut` everything else should be safe.

On preview, I agree with @roblabla again, though I think we still see differently about why Sync is an unsafe trait!

Rahix:
@korken89, I don't have enough time to write up my thoughts right now, I'll hopefully get around to it within the next week.  Though I still see some issues with the design regarding actually sharing such a mutex.  I'll provide the details later, once I got time to write it up.

perlindgren:
@roblabla, I now see the point you (all) are making and the arguments for making it safe to implement.

korken89:
I agree here (not unsafe), I'll update the text.

rfcs/0377-mutex-trait.md
```diff
@@ -0,0 +1,278 @@
...
+pub trait Mutex {
+    /// Data protected by the mutex
+    type Data;
+
+    /// Creates a critical section and grants temporary access to the protected data
+    fn lock<R>(&mut self, f: impl FnOnce(&mut Self::Data) -> R) -> R;
+}
+```
+
+The design is such that the data protected by the mutex is only accessible within the closure provided to the `lock` method, moreover the closure shall run within a critical section such that it is guaranteed that no-one else can access the same data at the same time. The critical section can for example be:
```
andre-richter:
```suggestion
The design is such that the data protected by the mutex is only accessible within the closure provided to the `lock` method. Moreover, the closure shall run within a critical section such that it is guaranteed that no-one else can access the same data at the same time. The critical section can for example be:
```

rfcs/0377-mutex-trait.md
```diff
@@ -0,0 +1,278 @@
...
+
+* Disabling interrupts (e.g. `cortex-m` / `risc-v` critical sections)
+* Analyzing the program and set BASEPRI in Cortex-M devices (Stack Resource Policy/RTFM way)
+* Analyzing the program and set interrupt masks correctly
+* etc
+
+## Helpers
+
+A concerned that has been raised is that taking multiple locks can lead to excessive rightward drift as:
```
andre-richter:
```suggestion
A concern that has been raised is that taking multiple locks can lead to excessive rightward drift as:
```

rfcs/0377-mutex-trait.md
```diff
@@ -0,0 +1,278 @@
...
+fn compile_error(mtx: &mut impl core_mutex::Mutex<Data = i32>) {
+    mtx.lock(|data| {
+        *data += 1;
+
+        mtx.lock(|data| *data += 1);
+    });
+}
+```
+
+Which produces the following error and making the mutex **deadlock-free by construction**:
```
andre-richter:
```suggestion
Which produces the following error and making the mutex **deadlock-safe by construction**:
```

rfcs/0377-mutex-trait.md
```diff
@@ -0,0 +1,278 @@
...
+
+The next thing is, **why should a lock be fallible**? One implementation of a spin-lock mutex was found which used a `Result` return to indicate a deadlock, but what other than a panic is correct if one would get a deadlock in a driver?
+
+All these questions point to an infallible `Mutex` trait. The example of a deadlock in a driver is to simply implement this trait and `.unwrap()` in the implementation.
+
+It boils down to this philosophy: `panic` is for program error - `Result` is for user error
+
+#### Should the trait be marked `unsafe`?
+
+The trait does not need to be marked as `unsafe` as the signature of `lock` together with the Rust type system guarantees exclusive access. If there is an implementation which handles multiple access points to the same mutable reference `unsafe` code must be used, which puts the responsibility of upholding Rusts alias rule on the implementor.
```
andre-richter:
```suggestion
The trait does not need to be marked as `unsafe` as the signature of `lock` together with the Rust type system guarantees exclusive access. If there is an implementation which handles multiple access points to the same mutable reference, `unsafe` code must be used, which puts the responsibility of upholding Rust's alias rule on the implementor.
```

rfcs/0377-mutex-trait.md
```diff
@@ -0,0 +1,278 @@
...
+```
+
+No. The `Mutex` trait is specifically for scoped use, i.e. the lock is defined in such a way that the resource protected by the lock is only accessible within the critical section that is the closure in the `lock` method.
+
+This means that for example if you have a peripheral driver and you want to lock access to it while it does its work, this mutex is not designed for that. However you can implement a higher level mutex on top of this one.
+
+> Should the `Mutex` trait not be in the [`embedded-hal`]?
+
+It is our strong recommendation to **not** place it in the [`embedded-hal`] crate.
+If there are breaking changes in the [`embedded-hal`] it can be so that competing implementations of the Mutex trait starts to exist which could bifurcate the ecosystem based on version, plus that the Mutex trait will be tied to HAL releases.
```
andre-richter:
```suggestion
If there are breaking changes in the [`embedded-hal`] it can be so that competing implementations of the Mutex trait start to exist which could bifurcate the ecosystem based on version, plus that the Mutex trait will be tied to HAL releases.
```

rfcs/0377-mutex-trait.md
```diff
@@ -0,0 +1,278 @@
...
+let x = lock1.lock();
+let y = lock2.lock();
+drop(x);
+drop(y);
+```
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+* Should the trait be unsafe to indicate that one needs to be aware of the design constraints to implement it?
```
andre-richter: Can be removed now that this is settled?

korken89: Yes, it should be removed - pushing that now

korken89:
Thanks for the nitpicks @andre-richter !

andre-richter:
While we are at it: Do we want to differentiate between mutex and spinlock. While they would have the same signature, it would make reading code more clear because it instantly tells what you are looking at.

korken89:
@andre-richter That should be an implementation of a Mutex, so its signature would be something like:

```rust
impl core_mutex::Mutex for Spinlock {
    // ...
}
```

I think this carries the intent quite well, or did you have something different in mind?

andre-richter:
I was thinking about functions that take the lock. They would have a signature taking an impl mutex, right?

That might be confusing and unclear What you actually get?

korken89:
I see what you mean. 
In my opinion that is a perfectly valid Mutex as it upholds the interface. Moreover I think we should be careful with bifurcating on types of Mutexes as this will reduce how generic a Mutex is.

It's up to the user to choose the Mutex of their choice :)

andre-richter:
When I raised my question I had in mind that in OS development world, the words `mutex` and `spinlock` unfortunately _usually_ have distinct meanings. `Mutex` implies that code will sleep through means of a scheduler when the lock is taken already, while a `spinlock` will busy wait.

So I was wondering if we could do something against the ambiguity that OS devs will face when they read `fn takes_lock(m: &mut impl Mutex<Data = >)`.

Some input from OS Devs would be nice. @roblabla or @phil-opp maybe?

Naive first idea: Naming the trait `core::lock` instead of `core::Mutex`.

perlindgren:
Could one think of extra traits for such properties? E.g., the RTFM Mutex could implement `NeverBlock` and `NoPanic`, meaning that `lock` is guaranteed to succeed immediately. 

Given those properties, we can design HAL drivers that are extremely efficient, robust etc. And still you are not bound to RTFM, any kernel/scheduler that offers a Mutex implementing `NeverBlock` and `NoPanic` would work.

roblabla:
Other potential names include `Lock` and `CriticalSection`. That said, [lock_api](https://docs.rs/lock_api/0.3.1/lock_api/trait.RawMutex.html) (which, BTW, might be a good case study as well) names its generic trait RawMutex, and the [spin](https://docs.rs/spin/0.5.2/spin/struct.Mutex.html) crate names its type Mutex. While I don't feel very strongly either way, I think keeping "Mutex" to stay coherent with the rest of the rust ecosystem is good.

I don't think we need different traits to define the properties? I think most code will either be written to be generic on the lock type, or use a specific lock implementation that has the precise semantics necessary for an operation. I wouldn't mind extra traits, I just wonder if they would actually get used.

andre-richter:
There is so much confusion with the taxonomy everywhere, its a shame. For example, take Wikipedia:
The article "Mutual Exclusion" lists [Types of mutual exclusion devices](https://en.wikipedia.org/wiki/Mutual_exclusion#Types_of_mutual_exclusion_devices).
One type being being a [Lock (mutex)](https://en.wikipedia.org/wiki/Lock_(computer_science)), which links back to the "Mutual Exclusion" article :confounded: 

 So Mutex can mean a principle and an implementation.

Maybe we should just put a disclaimer about the taxonomy Rust Embedded agreed on into the docs to settle this.

I would propose the following:
- "MUTual EXclusion" aka `Mutex` as used in this trait describes the _working principle_ and **not** the implementation characteristics. More specifically: 
  - A struct `Spinlock` could be one implementor of `trait Mutex` where the struct name gives away HOW mutual exclusion is achieved.
  - A struct `Sleeplock` would be another implementor with a telling name.
    - For unfortunate reasons, a sleeping lock is often also called just `Mutex` in existing code (most prominent example being Linux). Using that naming principle would mean a struct `Mutex` implementing the `trait Mutex`. This is discourage to resolve stated ambiguity.

nickray:
Outside comment: I'm in favour of reducing "weird words", especially if they even have conflicting usage in various fields by the experts. So I would vote for using `Lock`, which is descriptive, can have sleep, spin/busy etc. implementations, and the docs can add comments like "in field X, this is traditionally called such, in field Y..." etc.

(Tangentially, I find "critical section" bad naming for newcomers, when "interrupt free" or similar would be descriptive).

korken89:
Lets discuss this a bit extra at the meeting today, after this I will propose that we take this RFC into the final commenting period.

rfcs/0377-mutex-trait.md
```diff
@@ -0,0 +1,285 @@
...
+* Analyzing the program and set interrupt masks correctly
+* etc
+
+### Taxonomy
+
+A "MUTual EXclusion" aka `Mutex` as used in this trait describes the **principle** and not the *implementation* characteristics. More specifically:
+
+* A struct `Spinlock` could be one implementor of trait `Mutex` where the struct name gives away HOW mutual exclusion is achieved.
+* A struct `Sleeplock` would be another implementor with a telling name.
+		* For unfortunate reasons, a sleeping lock is often also called just `Mutex` in existing code (most prominent example being Linux). Using that naming scheme would mean a struct `Mutex` implementing the trait `Mutex` - this is discouraged to avoid stated ambiguity.
```
andre-richter:
```suggestion
  * For unfortunate reasons, a sleeping lock is often also called just `Mutex` in existing code (most prominent example being Linux). Using that naming scheme would mean a struct `Mutex` implementing the trait `Mutex` - this is discouraged to avoid stated ambiguity.
```

this still seems to render wrong.

Rahix:
I still have some concerns with the current design.  Specifically from the view-point of a library author who would want to use this trait in [`shared-bus`](https://crates.io/crates/shared-bus) ([intro post](https://blog.rahix.de/001-shared-bus/)).

Sorry if this is a bit messy ... I was kind of in a rush to get my thougths written down because I think there are some important points which still need to be discussed.  If any of this does not make sense I'd be happy to elaborate.

### 1. std-Mutex as part of the examples
In *How We Teach This*, two reference implementations based on `cortex-m` and RTFM are proposed.  I'd suggest a third one based on `std` as well.  That way, all the big different kinds of mutexes are covered.

### 2. Associated Type vs. Type-Parameter
I personally feel like a `Mutex<T>` is more consise to write and use than a `Mutex<Data = T>`.  It is also more in line with the rest of the code we are dealing with.  The implementing types will have to look like `SomeMutex<T>` anyway.

### 3. Use-cases
The *Motivation* section is very vague about the uses of such a mutex trait in my opinion.  It mentions applications but neither drivers nor other kinds of libraries.  I think the latter two are very big use-cases for generic mutexes.  Certain libraries and drivers *need* internal synchronization of state & hardware access between multiple independent accessors.  To name a few examples:

- [`shared-bus`](https://crates.io/crates/shared-bus) needs a mutex to synchronize bus accesses between independent drivers.
- @eldruin had mentioned using a mutex in their [`pcf857x`](https://github.com/eldruin/pcf857x-rs) driver to allow exposing port-expanded pins using the [`OutputPin`](https://docs.rs/embedded-hal/0.2.3/embedded_hal/digital/v2/trait.OutputPin.html) trait.
- In a [PR](https://github.com/Rahix/shared-bus/pull/6) to `shared-bus`, @geomatsi proposed sharing an ADC so the channels can be used independently.  This also needs a mutex.

To do these things there **must** be a generic mutex trait, as these libraries cannot know about concrete implementations.  In `shared-bus` I was forced to add my own mutex trait and implementations for it based on different HALs because of the current lack of a standardized mutex trait.

### 4. Mutex Creation
In my original proposal I have a `create()` method in the trait.  It looks like this:

```rust
fn create(v: T) -> Self;
```

In the RFC this topic is not discussed at all.  But from a library perspective, a generic way to create a mutex is very important.  Otherwise your library has to 'leak' details about the (possibly private) data it needs to store in the mutex.  In my opinion, this would lead to really inconvenient API designs.  Consider the following example:

```rust
struct PrivateData(u32, u32);

struct MyDriver<M: Mutex<PrivateData>> {
    privmutex: M,
}

impl<M: Mutex<PrivateData>> MyDriver<M> {
    pub fn new() -> MyDriver<M> {
        let privdata = PrivateData(0, 1);

        // How can the driver create a mutex here?
        let privmutex = ???;

        MyDriver {
            privmutex: privmutex,
        }
    }

    pub fn foo(&mut self) {
        self.privdata.lock(|p| {
            p.0 += 1;
            p.1 *= 2;
        });
    }
}
```

Back when writing `shared-bus` I had faced this issue and decided to go with the `create()` method to make the crate easier to use and expose less internal details.

Supporting a `create()` method is no issue with the following implementations:

- `std`: See the [relevant code](https://github.com/Rahix/shared-bus/blob/master/src/mutex.rs#L42-L51) in `shared-bus`.
- `bare-metal`-based, for example `cortex-m`: Again, see [here](https://github.com/Rahix/shared-bus/blob/master/src/mutex.rs#L54-L65).
- A possible implementation ontop of [freertos.rs](https://crates.io/crates/freertos_rs)'s [`MutexImpl`](https://docs.rs/freertos_rs/0.3.0/freertos_rs/struct.MutexImpl.html).

There is one sticking out where I don't see the possibility to implement a `create()` method:  *RTFM*.  I will come back to this later.

### 5. Sharing
I see difficulties with the current design regarding sharing mutexes.  Consider the following driver:

```rust
struct Driver<M: Mutex<Data = u32>> {
    m: M,
}

impl<M: Mutex<Data = u32>> Driver<M> {
    pub fn new(m: M) -> Driver<M> {
        Driver {
            m
        }
    }

    pub fn foo(&mut self) {
        self.m.lock(|d| {
            *d += 1;
            println!("D: {}", d);
        })
    }
}
```

Suppose you want to create two instance of this driver now, sharing access to the same peripheral.  You'd somehow need two mutable references to the mutex or two separate objects which you can give each instance of the driver.

`std` solves this using an `Rc`/`Arc`.  This is unapplicable for us A) because we don't have an `Rc` like data-structure in a lot of cases and B) because an `Rc` can only hand out immutable references while our mutex requires `&mut self`.

In the `shared-bus` implementation I kind of circumvented this problem by using lifetimes and immutable references; by having the mutex take `&self`, you can have as many references as you want:

```rust
impl<'a, M: 'a + Mutex<Data = u32>> DriverTakingReference<'a, M> {
    pub fn new(m: &'a M) -> DriverTakingReference<'a, M> {
        ...
    }
}

fn main() {
    let m = shared_bus::Mutex::create(123);

    let mut drv1 = DriverTakingReference::new(&m);
    drv1.foo();

    let mut drv2 = DriverTakingReference::new(&m);
    drv2.foo();
}
```

The RFC proposes implementing the mutex trait for a references to the mutex type, which would allow something similar to:

```rust
fn main() {
    let m = some::Mutex::new(123);

    Mutex::lock(&mut &m, |d| {
        ...
    });

    // &some::Mutex<T> implements the Mutex trait so this works:
    let mut drv1 = Driver::new(&m);
    drv1.foo();

    let mut drv2 = Driver::new(&m);
    drv2.foo();
}
```

Note how this looks similar to the example before, but is fundamentally different.  The driver does not need to know about the reference in the latter case.  I have also created a [playground to demo this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=b0f30ffc9a21271cd8b40c14e0b8e155).

**But**, this approach has a few issues:

1. > This implementation does however **allow for relocking the mutex within a lock again, so why bother** [...]?

   The RFC dismisses this approach entirely because of it not having any deadlock prevention.  At the moment, it does not mention any alternative ways to share a mutex though.
2. For an application writer, having to deal with `&mut &_` is going to be very unintuitive.  It makes onboarding very hard for people not too well acquainted with Rust's type system.
3. While it works in the above case, trying to do the same thing generically will be difficult.  The following code works, but depends heavily on the HALs providing all necessary implementations:

   ```rust
   fn two_drivers(m: impl Clone + Mutex<Data = u32>) {
       let mut drv1 = Driver::new(m.clone());
       drv1.foo();

       let mut drv2 = Driver::new(m);
       drv2.foo();
   }
   ```

   [Full playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=defa4afa510065f75668b11034ff52e2).
4. It requires work on the implementors side.  A HAL which does not implement the mutex trait for references will be incompatible with code trying to share a mutex.  Another example would be with the `std`-mutex.  Its implementation would also need an `impl<T> Mutex for sync::Arc<sync::Mutex<T>>` to allow sharing.  I thought about the possibility of a blanket implementation using `Deref` but I do not think that this is possible (because of `&mut self` vs `Deref` returning `&T`).

   [Playground showcasing the `Arc<Mutex<T>>`](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=a2c24d1071c3251f344eedbb424ba112).

### 6. RTFM & Types of Mutexes
I feel like the RFC in its current form is heavily based around compatibility with RTFM.  While it certainly is a great framework, I think it is very important to look beyond and have `embedded-hal` support other kinds of program architectures as well.

As seen with the above two points (4: `create()` method & 5: sharing), the RTFM mutex works *very* differently from any of the other mutex implementations and this introduces difficulties trying to bring all of them into one common trait.  I think this might not be the right approach after all.  To show what I mean, I'll try to highlight the differences:

Because the names in this are are all very loosely defined, I'll call the two **Type A Mutex** and **Type B Mutex** (If there are proper names for these, please tell me!).

#### Type A Mutexes
I'll call *Type A Mutex* the kind of mutex which is seen for example in the standard library: [`std::sync::Mutex`](https://doc.rust-lang.org/std/sync/struct.Mutex.html).  Its design is based on the idea of interior mutability; you can have *many references* to this mutex from many different threads/tasks/places because all those references are immutable.  You can even *duplicate* your own reference and give one copy to someone else if you want or move it across a thread boundary, etc.

#### Type B Mutexes
On the other hand, I'll call a *Type B Mutex* the kind of mutex RTFM uses.  Here, copies of the mutex are a highly policed resource.  You can **not** duplicate your copy, you can only temporarily borrow it to someone else.  Also, existing copies can **not** travel across task/thread boundaries.  So to speak, each 'copy' of the mutex is *unique*.  Because of this, everyone has mutable access to their 'copy' (which allows for the `&mut self` trick against deadlocks by recursive use).  By restricting the creation of such *Type B Mutexes* to procedural-macros, this allows the amazing compile-time guarantees RTFM can give, but obviously disallows dynamic creation of mutexes (like the `create()` method).

---

Loosely speaking, I'd argue that *Type B Mutexes* are a subset of *Type A Mutexes*.  You can implement a *Type B Mutex* by wrapping a *Type A Mutex* but I don't see a possibility to do this the other way around.

To show some code, here is a *Type B Mutex* based on `std::sync::Mutex`.  To initially create multiple 'copies' of the *Type B Mutex* I created a 'governor'.  Not that this governor is **not** part of the *Type B Mutex*.

**Implementation of a _Type B Mutex_ based on `std::sync::Mutex`**

```rust
// std::sync::Mutex is a Type A Mutex

// Type B Mutex based on std::sync::Mutex
// --------------------------------------

// Governor polices giving out copies
// Done by procedural-macro in RTFM
struct BMutexGovernor<T>(sync::Arc<sync::Mutex<T>>);

impl<T> BMutexGovernor<T> {
    pub fn new(d: T) -> BMutexGovernor<T> { ... }

    pub fn give_me_a_b_mutex(&self) -> BMutex<T> { ...}
}

struct BMutex<T>(sync::Arc<sync::Mutex<T>>);

impl<T> BMutex<T> {
    // **NO** BMutex::new() !!!

    // takes &mut self, because each 'copy' is unique and not a reference
    pub fn lock<R>(&mut self, f: impl FnOnce(&mut T) -> R) -> R { ... }
}

fn main() {
    let gov = BMutexGovernor::new(123);

    let mut m1 = gov.give_me_a_b_mutex();
    let mut m2 = gov.give_me_a_b_mutex();

    // Get rid of governor, now there is no way to get a new copy of the mutex
    drop(gov);

    m1.lock(|d| {
        println!("Got the lock! Value: {}", d);
        *d *= 2;

        // Deadlocking is not possible because of &mut self
    });

    m2.lock(|d| { println!("Got the lock from m2! Value: {}", d); });

    // There is no way to duplicate the copies
    // m2.clone();
}
```

I have created a [full playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=1f15fcd5d5ccb92c2bf97251035bde4a) to demo this.

Anyway, what I am trying to get towards is that I think we need to very explicitly make this distinction.  For many libraryies it will be very hard or impossible to work with a *Type B Mutex* while a big part of RTFM would break apart when introducing *Type A Mutexes* (at least from my limited understanding ... Maybe someone with deeper knowledge can comment on this?).  In my opinion it would be wrong to decide on one or the other approach as both clearly have their right to exist and we need to think about a good way to allow both possibilities without creating a big mess ...

---

I hope I was able to get my concern across ... With the current draft, I am pretty certain that it would be impossible to build crates like `shared-bus`.  But I also recognize how my original idea is completely incompatible with RTFM.  In my opinion, the only future-proof solution is one which can handle both cases elegantly ...

eldruin:
Thank you for the great write up @Rahix! I do share your concerns and have stopped short of implementing stuff like in the pcf857x crate in other drivers because of the lack of an usable `Mutex`.

korken89:
Thanks for the concerns @Rahix! 
I will do a write-up in the weekend, preliminary I think all your concerns can be handled under the current proposal.

korken89:
As the overall reservation is on how a shared bus implementation works with this trait, we have provided a quick-and-dirty proof of concept implementation that can be used as reference. One can probably implement a shared bus in different ways, but this should highlight that there is no problem in using the current trait.

[Link to implementation](https://github.com/korken89/core-mutex/blob/master/examples/shared_bus.rs)

Bellow follows more details answers.

---

> **1. std-Mutex as part of the examples**

Here I agree, an implementation can be provided without issue.

> **2. Associated Type vs. Type-Parameter**

The difference between these two is not just syntax. The latter allows for overloading the implementation and this can lead to inference errors. For example,

``` rust
trait Mutex<T> {
    fn lock<R>(&mut self, _: impl FnOnce(&mut T) -> R) -> R;
}

struct X;

impl Mutex<u8> for X {
    fn lock<R>(&mut self, _: impl FnOnce(&mut u8) -> R) -> R {
        unimplemented!()
    }
}

impl Mutex<u16> for X {
    fn lock<R>(&mut self, _: impl FnOnce(&mut u16) -> R) -> R {
        unimplemented!()
    }
}

fn error(mut x: X) {
    x.lock(|_| {});
    //~^ error: type annotations needed for `&mut T`

    // this does *not* pick the `Mutex` implementation; `u8` is the return type
    x.lock::<u8>(|_| {});

    // this works but it's verbose
    <X as Mutex<u8>>::lock(&mut x, |_| {});

    // this also works
    x.lock(|_: &mut u8| {});
}
```

If we want to make using `Mutex` less verbose we can rename the associated type to `T`. That's the name we use in `rtfm_core::Mutex`.

``` rust
fn foo(x: impl Mutex<T = u8>) { /* .. */ }
```

> **3. Use-cases**

Adding `shared-bus` as a use-case link is no issue

> **4. Mutex Creation**

> Otherwise your library has to 'leak' details

All libraries that use a type generic over a `Mutex` will run into this leaking problem. Using your example,

``` rust
// driver crate
pub struct PrivateData(u32, u32);
// ^ must be public or application crate won't be able to name the driver type

pub struct MyDriver<M: Mutex<Data = PrivateData>> {
    mutex: M,
}

// application crate
type AppDriver = MyDriver<spinlock::Mutex<PrivateData>>;
                                       // ^^^^^^^^^^ leaks "private" type

// or
type AppDriver = MyDriver<std::sync::Mutex<PrivateData>>;
                                        // ^^^^^^^^^^
```

The only way to truly hide `PrivateData` is to provide new types for each mutex implementation in the driver crate.

``` rust
// driver crate
struct PrivateData(u32, u32); // hidden

// exposed but can't be used due to hidden PrivateData type
pub struct MyDriver<M: Mutex<Data = PrivateData>> {
    mutex: M,
}

// exposed
pub struct SpinDriver(MyDriver<spinlock::Mutex<PrivateData>>);
```

---

> In my original proposal I have a create() method in the trait.

Adding this constructor to the trait makes it *less* flexible. Instead of adding a new method a separate trait can be provided then drivers can provide "from scratch" constructors *and* "from parts" constructors. For example,

``` rust
// mutex-core
// (object-safe trait: you can create `dyn Mutex` trait objects)
trait Mutex {
    type T;

    fn lock<R>(&mut self, _: impl FnOnce(&mut Self::T) -> R);
}

trait MutexCreate: Mutex {
    fn create() -> Self;
}

// driver crate
struct Something;

struct Driver<M>
where
    M: Mutex<T = Something>,
{
    m: M,
}

impl<M> Driver<M>
where
    M: Mutex<T = Something>,
{
    fn from_parts(m: M) -> Self {
        // ..
    }

    fn from_scratch() -> Self
    where
        M: MutexCreate,
    {
        // ..
    }
}
```

> **5. Sharing**

As you pointed out, not all mutexes can be freely created, the RTFM mutex being a good example of this. When it comes to making drivers which store the mutex I'd propose the following driver design which is compatible with both approaches:

```rust
// Some data used by the driver
struct DriverData;

// This is used with mutexes which can be used with reference or making copies of the Mutex
// Many will work but will not be compatible with for example RTFM
struct WrapperWithMutex<T> {
    mtx: T,
    driver: GenericDriver,
}

impl<T> WrapperWithMutex<T>
where
    T: core_mutex::Mutex<Data = DriverData>,
{
    fn new(mtx: T) -> Self {
        Self {
            mtx,
            driver: GenericDriver::new(),
        }
    }

    // Wrapped version
    pub fn do_something(&mut self) {
        self.mtx.lock(|data| {
            // Do something
        });
    }
}

// This driver is generic to the Mutex implementation
struct GenericDriver {
    // Fields ...
}

impl GenericDriver {
    pub fn new() -> Self {
        GenericDriver {
         // Fields ...
        }
    }

    // Here any Mutex can be used, std, RTFM, cortex-m
    pub fn do_something(mtx: &mut impl core_mutex::Mutex<Data = DriverData>) {
        mtx.lock(|data| {
            // Do something
        });
    }
}
```

Drivers written with the `GenericDriver` structure will work with the currently proposed mutex trait. However if you want to keep a mutex (or reference to it) within the driver structure you can create a wrapping driver which re-exports the `GenericDriver` methods as shown above. However this will only allow for a subset of mutexes (the one that can be created / stored) to be used with the `WrappingWithMutex`. But users not having this mutex can still use the `GenericDriver` without problem. 

> The RFC dismisses this approach entirely because of it not having any deadlock prevention. At the moment, it does not mention any alternative ways to share a mutex though.

I'm not sure I follow here, we have the example implementation which specifically implements for this case. The only difference is that this kind of mutex does not have the same compile time guarantees as a mutex not based on an `&mut &_` implementation, and run-time checks will do the work instead.

> For an application writer, having to deal with `&mut &_` is going to be very unintuitive. It makes onboarding very hard for people not too well acquainted with Rust's type system.

While it is not as straight forward, it is what Rust's type-system gives us to solve this problem. And this pattern can already be found in the standard library with for example `Rc` or `Arc`, any shared reference is a pointer and having a mutable reference to it is the same as the `&mut &_` pattern but under disguise.

To levitate this one can implement a `Ref` struct which simply wraps the reference, giving the interface `&mut Ref` which could be easier to swallow.

> **6. RTFM & Types of Mutexes**

The aim of the RFC is to be compatible with **all** mutexes, including RTFM, which the RFC accomplishes. It's important to see that this mutex is a very low level mutex, and to create more powerful locking mechanisms one needs to build on top of this core.

> Type A vs. Type B mutex

From the driver perspective one should not need to see a difference between the type of mutex, remember that this RFC aim to provide an interface to lock a mutex **not** an implementation and it is very important to see the difference here. And using the `&mut &_`/`&mut Ref` elegantly solves this for existing and new approaches.

korken89:
_Note from WG meeting on the 24/9:_
Wait 1 more week for comments.

Rahix:
Sorry for replying this late ...

> As the overall reservation is on how a shared bus implementation works with this trait, we have provided a quick-and-dirty proof of concept implementation that can be used as reference. One can probably implement a shared bus in different ways, but this should highlight that there is no problem in using the current trait.

Yes, the idea you outlined in the code is what I had thought about as well.  Though this only works for a `shared-bus`-type scenario where we can let the user create the mutex for us.  In the "private-data"-case this would not work.  But maybe we should try and gather some actual examples of drivers/libraries where this is the case to gauge whether this is actually an issue worth thinking about.

Another thing I'd like to talk about is the RTFM example you provided here.  From my understanding of RTFM it is not realistic to create the driver instances in the tasks as that will mean it is recreated every time the task runs.  Instead we'd want an implementation where the instances are created in `init()`.  This would also be more flexible in regards to cases where access to the same driver from multiple tasks is needed.

I think such an implementation is not possible with a design close to the current `shared-bus`.  I'd not worry about this too much though and keep `shared-bus` for use with *Type A Mutexes* / use-cases outside RTFM.  For RTFM we can think about a different bus-sharing design.  I have some ideas here but haven't tried anything yet.

> Adding this constructor to the trait makes it *less* flexible.

I agree, adding the `create()` method to the main `Mutex`-trait would be a very bad idea.  I like your

```rust
trait MutexCreate: Mutex {
    fn create() -> Self;
}
```

and would argue this should also be a part of `core-mutex`.

> The latter allows for overloading the implementation and this can lead to inference errors.

Oh, I had completely overlooked this!  In that case I agree, the associated type makes way more sense.  Would you mind detailing this design decision in the RFC?

> To levitate this one can implement a Ref struct which simply wraps the reference, giving the interface &mut Ref which could be easier to swallow.

I'd be much more comfortable with something going in this direction because the `Ref`-wrapper could protect against recursive use again by not implementing `Copy`.  What about adding this type to `core-mutex` as well (the name might not be ideal)?

```rust
#[derive(Clone, Debug)]
struct MutexRef<M: Mutex + Clone>(M);

impl<M: Mutex + Clone> Mutex for MutexRef<M> {
    type Data = M::Data;

    fn lock<R>(&mut self, f: impl FnOnce(&mut Self::Data) -> R) -> R {
        self.0.lock(f)
    }
}

impl<M: Mutex + Clone + MutexCreate> MutexCreate for MutexRef<M> {
    fn create() -> Self {
        MutexRef(M::create())
    }
}
```

> remember that this RFC aim to provide an interface to lock a mutex **not** an implementation

The *Type A* vs *Type B* stuff was meant to be about the interface the mutex provides, not its implementation.  The point where the two meet is where some implementations are not compatible with some mutex APIs.  Which is totally fine.  I would just like to see all capabilities exposed for mutex implementations which have them.  And in my opinion this needs to be part of `core-mutex` to prevent fragmentation of the eco-system.

---

My suggestion is this:

* The current mutex trait is fine as is
* Adding a `MutexCreate` trait for implementations supporting this
* Also adding a `MutexRef` type for shareable mutexes

This would allow mapping like this:

* A *Type A Mutex* is any type implementing `Mutex<Data = T> + MutexCreate + Clone`
* A *Type B Mutex* is any type implementing `Mutex<Data = T> + !MutexCreate + !Clone`

roblabla:
> I agree, adding the create() method to the main Mutex-trait would be a very bad idea. I like your MutexCreate suggestion

Uuh, shouldn’t this take a Murex::Data? I expected something more like

```
trait MutexCreate: Mutex {
    fn create(data: Self::Data) -> Self;
}
```

Otherwise the mutex wouldn’t get initialized. But this begs the question, why not simply use `From<Mutex::Data>`? Basically say that mutexes should implement From in order to allow easy generic construction.

IMO MutexCreate shouldn’t exist if it’s going to be a seperate trait. It’s redundant, Rust already has a plethora of construction traits.

eldruin:
I have an example that might be of interest here.
I think for some drivers like [pcf857x](https://github.com/eldruin/pcf857x-rs) (I/O expander) the [MutexI2c](https://github.com/korken89/core-mutex/blob/master/examples/shared_bus.rs) would work, and furthermore, allow for the user to decide whether he needs the I2C accesses to be mutexed and pay the performance penalty or not, which offers great flexibility.

However, this would not fit as well when multiple steps are necessary for an operation like in the driver for [xca9548a](https://github.com/eldruin/xca9548a-rs) (hardware I2C switch/multiplexer, useful for setups where several slaves share an address, for example).
I recently implemented splitting for this device so that several virtual I2C devices are created and can be passed on to other drivers.
```rust
let parts = i2c_switch.split();
let mut my_driver = Driver::new(parts.i2c0);
let mut my_other_driver = Driver::new(parts.i2c1);
```
When interacting with these virtual devices, before doing the actual `i2c.write`, it is necessary to select the appropriate channel by communicating with the switch device. See [here](https://github.com/eldruin/xca9548a-rs/blob/5a5bb1750722cc6aec07a55c8377569d3c076a67/src/parts.rs#L42).
Passing a `MutexI2c` as input to this driver would adequately lock the access to the `i2c` resource, which is good, but leave a race condition between the channel selection and the actual write.
This can lead to a situation where we communicate with the wrong device.
On single-threaded applications this is all solved by a `RefCell` as all the operations are done together in a closure after acquiring a mutable borrow. On multi-threaded applications, however, an additional mutex is necessary at the same level of the current `RefCell`.

What would be the best way to provide the user with the freedom to say whether accesses need to be mutexed or not and if so, provide the concrete type for the `Mutex`?

Rahix:
> Uuh, shouldn’t this take a Murex::Data?

@roblabla: Wow, yes, how did I miss this?!

> But this begs the question, why not simply use `From<Mutex::Data>`? Basically say that mutexes should implement From in order to allow easy generic construction.

Makes sense.  Though I think adding a `From<Mutex::Data>` bound will look confusing for people who don't know all the details so I'd suggest also adding a 'trait-alias' to make it easy to see the intent:

```rust
trait MutexCreate: Mutex + From<Mutex::Data> { }

impl<M: Mutex + From<Mutex::Data>> MutexCreate for M { }
```

Rahix:
@eldruin: We have the same issue in `shared-bus` with SPI right now, where there is a race-condition between setting the chip-select and locking the bus to do the transaction (See Rahix/shared-bus#8).

rfcs/0377-mutex-trait.md
```diff
@@ -0,0 +1,285 @@
...
+
+```rust
+/// Any object implementing this trait guarantees exclusive access to the data contained
+/// within the mutex for the duration of the lock.
+pub trait Mutex {
+    /// Data protected by the mutex
+    type Data;
+
+    /// Creates a critical section and grants temporary access to the protected data
+    fn lock<R>(&mut self, f: impl FnOnce(&mut Self::Data) -> R) -> R;
```
glandium: I must be missing something, but if you have a `&mut self`, why do you need a Mutex in the first place? In the `std::sync::Mutex` case, holding a `&mut self` allows to get a (wrapped) `&mut T` without a lock actually taking place.

korken89:
Hi all,

I'd like to thank all for all their input into this issue! We are going to start FCP now (1 week long), with the following:

* The RFC will have a minor update to comments and unless there are issues will be accepted as is
* The additions recommended by @Rahix and others that has been extensively discusses will be moved into the proposed crate - this to not need to go through voting again with this RFC (most discussed features only needs to go through official approval)

Thanks again for enlightening discussions!

Rahix:
Something we have not yet talked about at all is that the current design does not provide any way to acquire a mutex in a non-blocking fashion.  This makes the mutex unsuitable for use in ISRs and similar contexts which are not allowed to block under any circumstances.  It would probably be a good idea to include a second lock method that immediately returns if the mutex cannot be acquired.  That is:

```rust
pub trait Mutex {
    type Data;

    fn lock<R>(&mut self, f: impl FnOnce(&mut Self::Data) -> R) -> R;

    // Attempt locking the mutex but return immediately if this is not possible
    fn try_lock<R>(&mut self, f: impl FnOnce(&mut Self::Data) -> R) -> nb::Result<R, ()>;
}
```

korken89:
@Rahix This is by design and ties back to the original fallible vs infallible discussion. An fallible mutex can be implemented on top of this, however this RFC is specifically for an infallible mutex as we discussed earlier.

In the continuation of this we will look into expanding these traits and implementations and then this direction should have its own RFC/issue together with a specific discussion on only this to not miss important things as can happen if we make one RFC too large.

When we have the base crate up I will make issues for each of the extra features and ideas here and link in each person connected to it so each extension can get its own discussion.
Thanks for bringing it up!

Rahix:
> This is by design and ties back to the original fallible vs infallible discussion. An fallible mutex can be implemented on top of this, however this RFC is specifically for an infallible mutex as we discussed earlier.

The way I understand the fallible vs. infallible discussion, an infallible mutex is just defined to always be able to acquire the mutex _eventually_.  There are no guarantees about when that will happen and there might very well be a considerable time between start of the locking attempt and actual acquisition.  That means, there might be a time where the caller is blocked, waiting on the current mutex holder to release it.

This behavior makes the `lock()` function unusable in contexts where blocking is absolutely forbidden.  For example, in an ISR or a FreeRTOS timer callback.  For such contexts, a function is needed that returns immediately if the mutex is currently blocked by someone else.  This does not mean that our mutex type was fallible, it just means the mutex is currently in use elsewhere and will be unblocked some (finite) time in the future.

---

As a parallel, take a look at the mutex implementation in the Linux kernel:  `mutex_lock()` is defined like

```c
void mutex_lock(struct mutex *lock);
```

The `void` return type means that when the function returns, the mutex is now under _all_ circumstances owned by the current task.  Thus, locking a mutex is infallible, just like it is with our mutex trait.  This does not, however, forbid `mutex_lock()` from rescheduling until the mutex is freed by the current owner.

But for IRQs, tasklets and other contexts where rescheduling or any other kind of blocking is not allowed, there is a second

```c
int mutex_trylock(struct mutex *lock);
```

function that returns immediately whether it was able to acquire the mutex or not.  This function is now safe for any kind of context.

---

> In the continuation of this we will look into expanding these traits

I fear that standardizing a trait and expanding it later will be very painful as we saw with the `OutputPin`.  We will need version-hacks again for backwards-compatibility and it will take a very long time until all implementations update to the new upstream.  While we can't guarantee our current design to be flawless in any way, we can at least resolve any discussions which are brought up before standardization (even if it means to decide on postponing the feature).

japaric:
> the current design does not provide any way to acquire a mutex in a non-blocking fashion

Can you say more about the use case you have in mind for this *generic*
`try_lock` API? The `Mutex` trait is for inter-operation of crates via generic
code. How would a driver crate use this `try_lock` API, for instance, and how
would it be consumed by an application? (I)

---

I think I should say at this point -- not to anyone in particular -- that a
trait is not a good way to "standardize an API". Library authors are free to
implement or not any particular trait (e.g. `try_lock`, as proposed,
semantically makes zero sense for RTFM resources); announcing a trait is not
going to make the ecosystem grow crates that implement it; etc.

Traits (not extension traits, though) are mainly a code reuse mechanism; they
let you for example combine any driver (of M) with any HAL impl (of N) without
having to write `M * N` implementations -- only `M + N` implementations need to
be written.

---

> use in ISRs and similar contexts which are not allowed to block under any circumstances

IMO, in this case one should look for and use a *particular* mutex that has the
desired properties (for example RTFM mutexes are non-blocking; the entry and
exit code always executes in constant time) or the desired API (e.g.
`lock_timeout`) rather than use generic code or pick a type based on the traits
it implements.

To give an example, under this `try_lock` proposal would it be OK to use
`SomeDriver<M> where M: Mutex` in ISR / must-not-block context? How do I know if
`SomeDriver` is internally using the `lock` API or the `try_lock` API? I can't
tell from the signature; I would have to look at the implementation and if I
make a decision based on that then I'm relying on implementation details that
may change at anytime because they are not subject to semver. (II)

From this example alone, I would not add `try_lock` to the `Mutex` trait but
rather create a separate trait. Also, I should add that
`SomeDriver<rtfm::Mutex>` would be a good fit for must-not-block context under
the main proposal which does not include the `try_lock` API.

> As a parallel, take a look at the mutex implementation in the Linux kernel

I don't think this is a good comparison. This Linux API is a concrete function /
API, not a generic API with many implementations -- there's no code reuse there;
the whole kernel uses the same mutex type. This is a bit like saying we should
create a trait `X` *only because* there's a concrete type `X` in the Rust
standard library -- we should instead consider what code reuse benefits come
from creating such trait.

> I fear that standardizing a trait and expanding it later will be very painful
> as we saw with the OutputPin.

Very different scenario. `OutputPin` v2 has deprecated v1. The hypothetical
`try_lock` is meant to live alongside the `lock` method. Adding a separate
`try_lock` trait requires no compatibility wrappers or bridge types.

---

This `try_lock` idea also seems somewhat close to an "async" mutex (see
`futures::lock::Mutex`). Such mutex is also non-blocking but you get a
user-space context switch (at least when using it in a tokio context) instead of
blocking behavior (e.g. thread::sleep). Probably not something you want to call
from a ISR / time-sensitive context but the concept of an async mutex should also be
evaluated as a potential *additional* trait -- it seems very unlikely that a
single type is going to implement `lock`, `try_lock` and `async_lock` at the
same time, or even any two of these methods. (III)

---

In conclusion, I don't think a `try_lock` method should be added to the `Mutex`
trait (see II). It may make sense to have a separate trait for `try_lock`
depending on how useful it is for code reuse (based on the answer to I). It may
also be worth to think about an `async_lock` trait but this is still far away
given the state of no_std async and `impl Trait` in trait methods. (see III)

jamesmunns:
Not 100% sure what @Rahix had in mind, but I can see a failable mutex being useful in two situations:

1. Non-critical "concurrent" tasks, perhaps with 
2. A multicore system

For point 1, I can see why RTFM is advantageous, and can be considered to "obsolete" these needs, but for simple use cases, it may be useful to occasionally synchronize data without caring too much if we can't always obtain the mutex. For example if there are two "tasks" which both want to access the I2C bus (one to read a temperature sensor, once a second, the other to read an O2 sensor, once every 100ms), it would be nice to be able to simply wrap this in a non-blocking mutex, and just skip update periods when necessary. This isn't suitable for hard realtime or timing critical use cases, but sometimes it's okay to skip a sensor update every now and then if the bus is busy.

For the second use case (multicore), I could see a shared memory buffer that is protected by some kind of spinlock or basic CAS primitive for the purposes of a sharing data between two systems. We don't want to necessary block one core while the other is writing data, particularly if one is just seeing if any messages are pending.

japaric:
> I can see a failable mutex being useful in two situations:

I agree with these use cases. There are certainly uses cases for `try_lock` mutexes as a concept. My question was about a `try_lock` *trait* in particular; how would a trait help with these use cases? What APIs benefit from being written in a generic fashion?

In these two examples I would just go to crates.io and shop for a mutex with the right semantics not caring much about the API (the "shape"; i.e. closure API vs guard-based API).

korken89:
There is no issue in having a `try_lock` if we decide to have a trait for it in the future, I think we are starting to focus on the wrong part now.
The `Mutex` trait is not to have all possible methods in it, but to get the bare essential part and then expand with more traits as we find them needed.

However every extension we propose should have its own RFC to get proper focus and discussion as this is a very fundamental core area which is very easy to get wrong if we do not thoroughly think it through.

thejpster:
So I think I've been convinced that any scenario requiring a failable mutex (like round-robin processing of co-operative tasks which greedily acquire resources and hold on to them beyond their run-time as some hardware operation - e.g. DMA - is on-going in the background) can be resolved by using some other type (e.g. an atomic counter) and a infallible mutex. I agree with @japaric that `try_lock` shouldn't be in this trait.

almindor:
For what it's worth I also think having a separate `Trait` for `try_lock` is the way to go. The main problem with that approach is fragmentation of implementations where I think a lot of them would not implement the "extensions traits" and cause compatibility issues accross devices/drivers.

I don't think there's a good solution to that however, even if there was a nice standard chances are people would still only implement the minimum enforced functionality so it becomes a question of what's considered required functionality for a Mutex in general.

glandium:
Can someone please answer https://github.com/rust-embedded/wg/pull/377#discussion_r333350387 ?

rfcs/0377-mutex-trait.md
andre-richter: Its explained in the this Readme's chapter named: **The [Rust standard library `Mutex`] uses `&self` for the `lock` method, why choose `&mut self`?**

andre-richter: Line 91

jamesmunns:
We've had a lot of discussion, and decided in the meeting this is ready to merge.

Thanks all!

bors r+

jamesmunns:
bors retry

PR #396, opened 2019-11-22
Add myself to the Cortex-M team

jonas-schievink:


therealprof:
@thejpster @ithinuel @korken89 One more team member approval would be needed here.

therealprof:
And we have a @rust-embedded/cortex-m majority of 3 out of 5. Welcome @jonas-schievink!

PR #400, opened 2019-11-28
Increase Matrix room visibility

Disasm:
It turns out, the link to the chat room is difficult to find for the new users. This PR makes it more visible.


Disasm:
Also it will be good to set URL for the `rust-embedded` org to point to the `https://github.com/rust-embedded/wg` repository.

PR #409, opened 2019-12-17
Add reorgansing-hal RFC

thejpster:
This RFC proposes a new mechanism for handling updates to the embedded-hal crate, through the use of unstable *sub-crates*.

[Rendered](https://github.com/thejpster/wg/blob/thejpster-reorg-hal-rfc/rfcs/000-reorgansing-hal.md)

### DIFF
--- a/rfcs/000-reorgansing-hal.md
+++ b/rfcs/000-reorgansing-hal.md
@@ -1,0 +1,64 @@
+- Feature Name: Reorganising the Embedded HAL
+- Start Date: 17-Dec-2019
+- RFC PR: 
+- Rust Issue: 
+
+# Summary
+[summary]: #summary
+
+Currently when the Embedded HAL wants to roll out new version of a trait, it does so by embedded the version number into
+the module name - e.g. `embedded_hal::digital::v2`. This approach exposes to much internal complexity to the end user and
+doesn't give us anywhere to experiment with new features. This RFC proposes creating sub-crates to replace the current
+modules, and replacing the embedded-hal crate with a simple crate that only re-exports stable types from the relevant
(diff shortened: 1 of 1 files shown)


almindor:
One thing to check for before it becomes a problem "mid-way" is if there's any possibility of the foreign-foreign problem with Traits.

I've recently been refactoring one of my Rust projects and when I tried to remove all serializable types out to a separate trait came to a case where I had to also push `impl OtherLib::Trait for Type` inside the types crate because I used another external library that enforced its Traits on the types to function.

In other words I had to push all external required traits to the lowest crate. I don't think we have this problem but given how ugly this is I thought it might be worth a check.

rubberduck203:
If we break `embedded-hal` into many focused crates, is there any longer a need for the `embedded-hal` crate at all?  Perhaps it is still useful for allowing individual traits to evolve independently of their implementations, then the new versions are pulled into the "uber" crate when  the wg determines a sufficient number of hals and drivers have been updated? 

Under drawbacks, I think we've missed the possible fragmentation that this may create. IMO, that's the biggest risk associated with this proposal. Although, it's a risk I believe to be worth while. Taking the approach I mentioned above minimizes this risk of fragmentation, I _think_. 

Another minor drawback will be the frequency of major versions of `embedded-hal`, although, this would also be a small issue with keeping the crate as is, and using semver instead of namespacing for versioning the "sub crates". Speaking of, is that an approach that has been thought out, just semvering the crate properly as is? I mentioned a few weeks ago at a wg meeting that it's typically best to use an "add, deprecate, remove" approach to evolving interfaces. At least a few people seemed to like the idea.

PR #419, opened 2020-01-26
Address Multi-Core Soundness by abolishing Send and Sync

jonas-schievink:
# Summary

Change how we model multi-core MCUs in the following ways:

* Stop using `Send` and `Sync` to model anything related to cross-core
  interactions since those traits are unfit for the purpose.
* Declare that `Send` is not sufficient to transfer resources between cores,
  since memory addresses can have different meanings on different cores.
* Mutexes based on critical sections become declared sound due to the above:
  They can only turn `Send` data `Sync`, but neither allows cross-core
  interactions anymore.
* Punt on how to deal with cross-core communication and data sharing for now,
  and leave it to the ecosystem.

[Rendered](https://github.com/jonas-schievink/wg/blob/multi-core-soundness/rfcs/0000-multi-core-soundness.md)

 [`bare-metal`]: https://github.com/rust-embedded/bare-metal

### DIFF
--- a/rfcs/0000-multi-core-soundness.md
+++ b/rfcs/0000-multi-core-soundness.md
@@ -12,12 +12,11 @@ Change how we model multi-core MCUs in the following ways:
   interactions since those traits are unfit for the purpose.
 * Declare that `Send` is not sufficient to transfer resources between cores,
   since memory addresses can have different meanings on different cores.
-* Introduce `CoreSend` and `CoreSync` traits to the [`bare-metal`] crate.
-* Require multi-core PACs/HALs, frameworks and/or libraries to manually mark
-  data that is safe to send across core boundaries.
 * Mutexes based on critical sections become declared sound due to the above:
   They can only turn `Send` data `Sync`, but neither allows cross-core
   interactions anymore.
+* Punt on how to deal with cross-core communication and data sharing for now,
+  and leave it to the ecosystem.
 
 # Motivation
 [motivation]: #motivation
@@ -102,59 +101,20 @@ risk of the resources changing their meaning when sent.
 # Detailed design
 [design]: #detailed-design
 
-## Introduce common traits for modeling cross-core operations
+## Document what `Send` and `Sync` mean
 
-The following trait definitions and impls will be added to [`bare-metal`]:
+`Send` and `Sync` will be used only to model transfer and sharing of resources
+between different *execution contexts* that run with the same fixed set of
+global resources.
 
-```rust
-/// Types that can be transferred across core boundaries.
-pub unsafe trait CoreSend {}
+Here, an *execution context* is something that may execute code asynchronously
+from other code (so without needing to be called). For example, a *thread* or
+an *interrupt handler* would qualify as an *execution context*, while a
+single-threaded futures executor would *not* create any more *execution
+contexts* while it executes.
 
-/// Types that can be accessed from multiple cores at once.
-pub unsafe trait CoreSync {}
-
-unsafe impl<'a, T: CoreSync> CoreSend for &'a T {}
-```
-
-Note that these are not auto traits. That is not just because auto traits are
-still an unstable feature, but also because that wouldn't be correct: As
-outlined in [Today's Soundness Issues][todays-issues], whether some data is safe
-to be sent between cores depends not just on its type contents, but also on its
-location in memory.
-
-It is only safe to implement `CoreSync` if *all* instances of the implementing
-type will either be at memory locations shared by all cores, or are zero-sized.
-Similarly, it is only safe to implement `CoreSend` if *all* instances of the
-implementing type contain only data that is safe to send to another core (which
-moves the value in memory). Of course, the type's API must also ensure that no
-memory accesses to core-local memory occur (important for eg. core-local
-peripherals).
-
-### `CoreSend`
-
-It is expected that `CoreSend` will be implemented for peripherals shared
-between all cores in the system, and also by HAL wrappers of those peripherals,
-so that applications can transfer such peripherals to any core in the system.
-This RFC does not mandate how exactly those implementations will be written or
-generated. One option would be to give `svd2rust` an unsafe command line flag to
-generate `CoreSend` for all peripherals.
-
-For example, `CoreSend` would be implemented by all peripherals of the LPC5411x,
-which has a Cortex-M4 and a Cortex-M0+ core sharing all non-core peripherals. It
-would *not* be implemented by the peripherals of the nRF5340, which has two
-cores that have independent peripherals (each core owns its peripherals).
-
-`CoreSend` is expected to be used as a trait bound by support libraries, HALs
-and frameworks such as [µAMP] that provide support for multi-core processors.
-For example, an API could be built that consumes `CoreSend` types and passes
-them to the entry point of the secondary core.
-
-### `CoreSync`
-
-The `CoreSync` trait is expected to be used for sharing *data* between cores.
-A framework like [µAMP] could generate a controlled wrapper type around shared
-data placed in a known section, and have that wrapper implement `CoreSync`,
-allowing the application to copy references to the data between cores.
+Concretely (for embedded Rust), that means `Send` and `Sync` can be used to
+model threads in an RTOS, or interrupt handlers for bare-metal applications.
 
 ## `bare_metal::Mutex` is now sound
 
@@ -175,34 +135,16 @@ In SMP apps, only a single executable is used for all cores, while each core can
 have a separate entry point (or another mechanism of identifying the running
 core). This means that all `static`s are shared by default.
 
-Since defining a `static` only requires that its type is `Sync`, not `CoreSync`,
-this would be unsound. For example, it would allow storing a `bare_metal::Mutex`
-in a `static` and access it from all cores. Therefore, this RFC foregoes the
-ability to write safe SMP apps in Rust, instead proposing to shift focus to AMP
-apps, which do not share data by default and produce a separate executable per
-core.
+Since defining a `static` only requires that its type is `Sync`, this would be
+unsound. For example, it would allow storing a `bare_metal::Mutex` in a `static`
+and access it from all cores. Therefore, this RFC foregoes the ability to write
+safe SMP apps in Rust, instead proposing to shift focus to AMP apps, which do
+not share data by default and produce a separate executable per core.
 
 # How We Teach This
 [how-we-teach-this]: #how-we-teach-this
 
-API documentation of `CoreSend` and `CoreSync` needs to be clarified. The
-single-sentence documentation above is not meant to be complete.
-
-## Document what `Send` and `Sync` mean
-
-`Send` and `Sync` will be used only to model transfer and sharing of resources
-between different *execution contexts* that run with the same fixed set of
-global resources.
-
-Here, an *execution context* is something that may execute code asynchronously
-from other code (so without needing to be called). For example, a *thread* or
-an *interrupt handler* would qualify as an *execution context*, while a
-single-threaded futures executor would *not* create any more *execution
-contexts* while it executes.
-
-Concretely (for embedded Rust), that means `Send` and `Sync` can be used to
-model threads in an RTOS, or interrupt handlers for bare-metal applications.
-
+(see above)
 
 # Drawbacks
 [drawbacks]: #drawbacks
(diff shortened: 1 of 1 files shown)


jamesmunns:
I’ll ask one possibly naive question:

Why do we need a separate model to represent multi-core behavior in bare metal programs, when this is not necessary for desktop CPUs with multiple cores? Is this due to the operating system abstracting away the reality of disjoint sets of resources between CPU cores in hosted environments?

For microcontrollers, we’ve taken the metaphor that “threads” can mean:

* The main thread of execution prompted by Main
* threads of execution originating from Interrupts/Exceptions, including RTFM’s tasks
* perhaps the operation of software threads, provided by an RTOS like Tock
* maaaaaybe the operations of DMA transactions, though those behave without code

In desktop PCs, there is no difference to Rust between moving data between threads on the same CPU/Core, or to threads on another CPU/Core.

Is this a model we should be aiming to upstream to Rust proper, perhaps as an uncommon but existent component of core? Or is it only relevant where the context details of actual execution matters, and should live within the embedded domain?

jonas-schievink:
> Why do we need a separate model to represent multi-core behavior in bare metal programs, when this is not necessary for desktop CPUs with multiple cores? Is this due to the operating system abstracting away the reality of disjoint sets of resources between CPU cores in hosted environments?

Correct. While APIs exist that allow making threads *not* share resources, and make processes *do* share resources, the "default" is that all threads within a program do share all of the program's resources, and this is the only use case intended to be modeled by `Send` and `Sync`.

> Is this a model we should be aiming to upstream to Rust proper, perhaps as an uncommon but existent component of core? Or is it only relevant where the context details of actual execution matters, and should live within the embedded domain?

For now, I don't think this is of much general interest. The proposed traits are very embedded-specific and I don't see how they can be made more general (nor do I think they should). While there's certainly a use case for making shared memory and IPC in general safe to use, that would likely require different traits specific to the exposed mechanism.

Amanieu:
I think that this RFC is missing a key section: what types will implement `CoreSend`/`CoreSync`, and how do you expect it to be used? Some examples would help here.

adamgreig:
I like this RFC; it keeps the common case (single core) simple and correct for existing and future code, we could still use Send and Sync with future RTOSs that could correctly handle multiple threads on one or multiple cores, and for today's multi-core programs, treating them as separate processes that need to use some form of IPC to communicate seems to map well to hosted Rust semantics.

My understanding is that bare-metal multi-core programs already have distinct entry points for each core, often with distinct binaries for each core too? So treating them as separate processes seems fine. We'll definitely want some sort of IPC for sharing data between those cores, but it doesn't have to be the same primitives we use for single-core synchronisation.

I think this RFC would also work without `CoreSync` and `CoreSend` at all; we could have one RFC to say "multi-core bare metal should be considered multiple processes for Send and Sync purposes" which fixes our current soundness issue and sets out a useful single-core foundation, and then a separate, more detailed RFC to introduce `CoreSync` and `CoreSend`, perhaps with some actual implementations of multi-core synchronisation.

jonas-schievink:
> I think that this RFC is missing a key section: what types will implement `CoreSend`/`CoreSync`, and how do you expect it to be used? Some examples would help here.

You're right, I've added some more text expanding on how they're expected to be used.

> My understanding is that bare-metal multi-core programs already have distinct entry points for each core, often with distinct binaries for each core too?

Yeah, you can either build a single image (SMP) or distinct images that can opt into sharing certain `static`s (AMP). The first case is unlikely to be supported though (mostly because it's really difficult to make `cortex_m::Peripherals::take()` work on all cores). I did add some text explaining my reasoning for why this proposal is at least *safe* when building an SMP app though.

> I think this RFC would also work without `CoreSync` and `CoreSend` at all

That's true. I've added that to the alternatives.

rfcs/0000-multi-core-soundness.md
```diff
@@ -0,0 +1,244 @@
...
+## Today's Soundness Issues
+[todays-issues]: #todays-soundness-issues
+
+The auto trait behavior of `Send` and `Sync` causes soundness issues even today:
+Imagine a user-defined struct `S` that only contains some primitive types.
+Clearly, this struct will and should automatically implement `Send` and `Sync`.
+Now, because `S` implements `Sync`, `&S` will implement `Send`. This is simply
+the languages definition of these traits and nothing we can change. Now a big
+problem can arise when tranferring `&S` across core boundaries: Cores can have
+private memory regions that are not accessible by other cores. If `T: Send` is
```
andre-richter: For the purpose of being able to fluently read this RFC, I would put one of the more popular cases as an example right after this sentence, or in braces.

rfcs/0000-multi-core-soundness.md
```diff
@@ -0,0 +1,244 @@
+- Feature Name: `multi_core_soundness`
+- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Change how we model multi-core MCUs in the following ways:
+
```
andre-richter: A notice here that this RFC is about Asymmetric MP and not SMP would be helpful to set the context before reading.

andre-richter:
> > Is this a model we should be aiming to upstream to Rust proper, perhaps as an uncommon but existent component of core? Or is it only relevant where the context details of actual execution matters, and should live within the embedded domain?
> 
> For now, I don't think this is of much general interest. The proposed traits are very embedded-specific and I don't see how they can be made more general (nor do I think they should). While there's certainly a use case for making shared memory and IPC in general safe to use, that would likely require different traits specific to the exposed mechanism.

While reading this, I had the same thought as @jamesmunns. One of THE key features of compiling a Rust program is that the compiler stops you from sharing or sending data across thread boundaries. Unless you resolve the issue that the compiler wants to see `Send`/``Sync` where it thinks it belongs, you won't get a binary.

The RFC addresses important issues, but right now, compiler enforcement will only kick in when the traits are set where they should be set. Users using provided libraries that implement what the RFC demands will be reminded when they use the API wrong. But it won't stop them to use Send/Sync wrongly in other places of the program.

Just a wild thought, but maybe a good companion to this RFC would be some global that "disables" Send/Sync for a specific memory region.

```rust
// Private memory regions live here
#![amp_range(0x8000-0x9000)]
```

jonas-schievink:
> The RFC addresses important issues, but right now, compiler enforcement will only kick in when the traits are set where they should be set. Users using provided libraries that implement what the RFC demands will be reminded when they use the API wrong. But it won't stop them to use Send/Sync wrongly in other places of the program.

I'm not completely sure what you're referring to here. Do you mean someone could write a library that provides an incorrect unsafe abstraction by taking any `T` instead of `T: Send`? Or that someone writes an incorrect `unsafe impl Send for X {}`?

> Just a wild thought, but maybe a good companion to this RFC would be some global that "disables" Send/Sync for a specific memory region.

The compiler doesn't know anything about where variables go, only the linker does (and we heavily rely on correctly written linker scripts even today). I'm not sure how this would work since those traits are implemented for types, not type-address-tuples. `Send` and `Sync` are just ordinary traits that can either be implemented or not. Do you suggest that the compiler shouldn't treat `static`s placed in private regions as implementing `Sync`?

rfcs/0000-multi-core-soundness.md
jonas-schievink: It is about both, since it is independent of the application model. Basically it just spells out a few facts about multi-core MCUs in general.

cr1901:
>Declare that Send is not sufficient to transfer resources between cores,
since memory addresses can have different meanings on different cores.

>Correct. While APIs exist that allow making threads not share resources, and make processes do share resources, the "default" is that all threads within a program do share all of the program's resources, and this is the only use case intended to be modeled by Send and Sync.

Would it be accurate to summarize the above two texts as the following?:

`Send` and `Sync` are only appropriate for multiple cores if code outside your application (be it a runtime or carefully implemented startup code/exception) contains a mechanism (such as an MMU or perhaps PIE) to make all globally scoped resources visible to each core.

andre-richter:
> The compiler doesn't know anything about where variables go, only the linker does (and we heavily rely on correctly written linker scripts even today). I'm not sure how this would work since those traits are implemented for types, not type-address-tuples. `Send` and `Sync` are just ordinary traits that can either be implemented or not. Do you suggest that the compiler shouldn't treat `static`s placed in private regions as implementing `Sync`?

Wasn't my most well-though out comment, true. Please disregard most of it.

What about creation of raw pointers. Could the compiler theoretically make use of the address used for casting?

jonas-schievink:
> What about creation of raw pointers. Could the compiler theoretically make use of the address used for casting?

Well it can always insert a runtime check, but so could any hypothetical multi-core MCU library with some sort of `SharedRef<'a, T>` wrapper that's `CoreSend` if `T: CoreSync`. That's precisely the sort of experimentation I'd like to see in ecosystem crates.

jonas-schievink:
> `Send` and `Sync` are only appropriate for multiple cores if code outside your application (be it a runtime or carefully implemented startup code/exception) contains a mechanism (such as an MMU or perhaps PIE) to make all globally scoped resources visible to each core.

That sounds about right, yeah. The mechanism can also choose to only make a subset of all global resources available, of course.

therealprof:
> My understanding is that bare-metal multi-core programs already have distinct entry points for each core, often with distinct binaries for each core too?

Not necessarily. I've learned from the SiFive guys that their RISC-V cores all start from the same entry point and the bootstrapping code running on each core is in charge of figuring out which core it currently is running on and act accordingly, e.g. by disabling itself or by jumping into the core specific code. I would presume there're quite a few implementations working like that.

therealprof:
@jonas-schievink 

Aren't

> Each core can have private peripherals that may be Sendable between interrupt handlers, but that do not exist (or, worse, something else exists at their address) on other cores of the system.

and

> Something much closer to a multi-threaded program is one that makes use of interrupt handlers: In both cases, all global resources exist precisely once, and may be shared and exchanged between interrupt handlers or threads (provided they implement Send/Sync) with no risk of the resources changing their meaning when sent.

saying the exact opposite?

jamesmunns:
@therealprof I believe in those quotes the former was describing multi-core, similar to disjoint processes; while the latter was describing a single core with interrupts, similar to disjoint threads.

jonas-schievink:
@therealprof What James said; I've pushed a commit that should clarify that.

rfcs/0000-multi-core-soundness.md
```diff
@@ -0,0 +1,244 @@
...
+## Interactions with Symmetric Multi-Processing (SMP) apps
+
+In SMP apps, only a single executable is used for all cores, while each core can
+have a separate entry point. This means that all `static`s are shared by
+default.
+
+Since defining a `static` only requires that its type is `Sync`, not `CoreSync`,
+this might seem like an unsound configuration. However, since all `static`s are
+in shared memory accessible by all cores, the `CoreSync` contract is fulfilled.
```
jonas-schievink: Note to self: This is incorrect, it allows sharing `bare_metal::Mutex` (which is `Sync`) between cores, which is unsound.

rfcs/0000-multi-core-soundness.md
```diff
@@ -184,6 +169,19 @@ support crates or frameworks like [µAMP].
...
 [`bare-metal`]: https://github.com/rust-embedded/bare-metal
 
+## The fate of Symmetric Multi-Processing (SMP) apps
+
+In SMP apps, only a single executable is used for all cores, while each core can
+have a separate entry point (or another mechanism of identifying the running
+core). This means that all `static`s are shared by default.
+
+Since defining a `static` only requires that its type is `Sync`, not `CoreSync`,
+this would be unsoung. For example, it would allow storing a `bare_metal::Mutex`
```
hannobraun:
```suggestion
this would be unsound. For example, it would allow storing a `bare_metal::Mutex`
```

Disasm:
I'd like to suggest considering the following cases:
* Different protection levels (TrustZone in ARM, M-S-U modes in RISC-V). These may limit the memory accessible as well as the subset of the peripherals. What markers should have these "protected" peripherals in this case?
* Different addressing modes (physical/virtual memory). Obviously, object cannot be Sync between different addressing modes even on the same core.
* The same firmware can be loaded into a core-local memory region as well as into a global memory. What markers should have objects, constructed by this firmware?

thejpster:
I'd personally rename `CoreSend`/`CoreSync` to `MultiCoreSend/MultiCoreSync`, but otherwise I'd take this over #388.

almindor:
> I'd like to suggest considering the following cases:
> 
> * Different protection levels (TrustZone in ARM, M-S-U modes in RISC-V). These may limit the memory accessible as well as the subset of the peripherals. What markers should have these "protected" peripherals in this case?
> * Different addressing modes (physical/virtual memory). Obviously, object cannot be Sync between different addressing modes even on the same core.
> * The same firmware can be loaded into a core-local memory region as well as into a global memory. What markers should have objects, constructed by this firmware?

It seems to me that ultimately Rust should support something like `MarkedPin<Marker, T>` which would require translation where possible (e.g. core1 local memory to core2 local memory if such case is permitted) and disallow any access otherwise. This needs further research tho the use cases and combinations are pretty varied.

jonas-schievink:
Updated the RFC to remove `CoreSend`/`CoreSync`, as per discussion in the meeting yesterday

jamesmunns:
As per our discussion yesterday, this RFC is now entering **Final Comment Period**, with the **disposition to merge**. As this issue has been open for more than two weeks already, I would like to merge this on **2020-02-18** (the date of our next meeting) if there are no blocking objections.

@rust-embedded/all, please provide a final review, and approve or reject this RFC.

jamesmunns:
bors r+

Thanks all!

Issue #266, opened 2018-12-10
Form a libs team?


japaric:
We've received quite a few crate requests in the 2019 wishlist survey. Those requests are listed below from most voted to least voted:

- [An easy to use USB device library](https://github.com/rust-embedded/wg/issues/256#issuecomment-438570134)
- [An easy to use USB host library](https://github.com/rust-embedded/wg/issues/256#issuecomment-438569808)
- [An easy to use pure-Rust RTOS](https://github.com/rust-embedded/wg/issues/256#issuecomment-438571019)
- [An easy to use SD/MMC and FAT32 library](https://github.com/rust-embedded/wg/issues/256#issuecomment-438569397)
- [High quality RTOS bindings](https://github.com/rust-embedded/wg/issues/256#issuecomment-438489820)
- [A task scheduler](https://github.com/rust-embedded/wg/issues/256#issuecomment-438935300)
- [GUI with desktop simulation support](https://github.com/rust-embedded/wg/issues/256#issuecomment-443406386)

The first two requests might fit within the scope of the HAL team but the rest don't fit within the scope of any of the other teams we currently have.

The question is: should we create a new libs team dedicated to cater to these requests?

My opinion is that we *should* create a libs team but should *not* add non-architecture specific (i.e. general purpose) crates to the org. Instead the libs team should follow the [lib blitz] model to help grow crates developed by the community.

[lib blitz]: https://blog.rust-lang.org/2017/05/05/libz-blitz.html

The reasons why I think we should, in general (\*), not maintain or develop general purpose no-std crates in the org are:

- Increased maintenance load (the WG currently owns 38 repos)

- Some crates require specialized knowledge (e.g. USB), which means we'll have to further increase the WG membership (the WG currently has 26 members)

- Trade-offs in library design often lead to competing implementations (e.g. RTOS-es), which is a good thing because people have different needs (e.g. some may prefer ease of use over flexibility) -- more options means they can pick what suits them best. Taking a crate into the org effectively blesses it (i.e. it puts a "WG approved" stamp on it). This, I think, would be rather harmful in the early stages of the embedded ecosystem as it would prevent other implementations from competing on equal footing.

(\*) of course, every rule has it exception. I think for example that, once mature, a core `usb` crate along the lines of the [`http`] crate would be reasonable to have under the org and the maintenance of the HAL team.

[`http`]: https://crates.io/crates/http

[...]

PR #431, opened 2020-02-25
Transfer smoltcp and rust-manged into WG under Tools team

adamgreig:
This PR would transfer [smoltcp](https://github.com/smoltcp-rs/smoltcp) and its dependency [rust-managed](https://github.com/smoltcp-rs/rust-managed) into the WG under the Tools team.

From their READMEs:
> _smoltcp_ is a standalone, event-driven TCP/IP stack that is designed for bare-metal, real-time systems

> _managed_ is a library that provides a way to logically own objects, whether or not heap allocation is available

smoltcp is looking for a new home and the current maintainers are happy to transfer it to the rust-embedded organisation, including push access to crates.io. The idea of transferring the crates was discussed in the weekly meetings on 2020-02-18 and 2020-02-25, with this PR being the outcome.

At the moment the Tools team seems the best fit for these libraries, but if/when #266 goes ahead it's anticipated we'd move these repositories to be under that new libs team. Right now none of the Tools team are necessarily very experienced with smoltcp maintenance; the expectation is that existing contributors will be able to continue working on smoltcp, while having it under the WG will help encourage new contributors and ensure continuity for this useful embedded crate.

Additionally there are some worries that having smoltcp under rust-embedded would be seen as blessing a single implementation to the detriment of other embedded TCP/IP stacks; to address this we intend to clearly note in the README that while smoltcp is maintained by the rust-embedded organisation, we don't intend for it to be the only embedded TCP/IP stack and encourage experimentation with alternatives. If multiple embedded TCP/IP stacks become widespread, we could seek to have abstraction/portability libraries in the WG and move the actual implementations to their own organisations.

This PR needs to be approved by the Tools team:

* [x] @adamgreig
* [ ] @burrbull
* [ ] @Disasm
* [x] @Emilgardis
* [ ] @reitermarkus
* [x] @ryankurte
* [x] @therealprof

### DIFF
--- a/README.md
+++ b/README.md
@@ -337,6 +337,8 @@ Projects maintained by the tools team
 - [`cargo-binutils`]
 - [`cross`]
 - [`itm`]
+- [`rust-managed`]
+- [`smoltcp`]
 - [`svd-parser`]
 - [`svd2rust`]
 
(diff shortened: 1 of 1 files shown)


adamgreig:
Ping  @rust-embedded/tools for feedback.

Disasm:
I like the idea in general, but I'd like to see how technically this maintenance under rust-embedded will be achieved. I think it's better to have at least two of the current maintainers as part of the tools team and to ensure that they are ok with the new rules (bors, etc).

adamgreig:
Closing since smoltcp is now maintained under its new smoltcp-rs organisation. We might revisit the idea of a libs team that could look after libraries like this in the future, though.

PR #435, opened 2020-04-04
Added hal-team meeting minutes

ryankurte:
so we have a canonical source, cc. @jamesmunns


ryankurte:
thanks @therealprof !

PR #441, opened 2020-04-09
Fix broken link

mendelt:
The link to the embedded rust book was broken.

### DIFF
--- a/README.md
+++ b/README.md
@@ -440,7 +440,7 @@ Our Matrix room is logged by [logbot] on the bridged IRC channel, and you can fi
 [@rustembedded twitter]: https://twitter.com/rustembedded
 [Awesome embedded Rust]: https://github.com/rust-embedded/awesome-embedded-rust
 [The Discovery book]: https://github.com/rust-embedded/discovery
-[The embedded Rust book]: https://rust-embedded.github.io/book/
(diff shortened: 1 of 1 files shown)


therealprof:
@mendelt Thanks for noticing. I think we should prefer https://doc.rust-lang.org/stable/embedded-book/ though.

mendelt:
> @mendelt Thanks for noticing. I think we should prefer https://doc.rust-lang.org/stable/embedded-book/ though.

No problem. I searched around a bit and there is also https://docs.rust-embedded.org/book/ which is already used on line 19 of the readme. Wouldn't that be better?

therealprof:
Oh, actually I got confused there for a bit. After revisiting the location where the broken link is, this one should actually point to the GitHub repository, so https://github.com/rust-embedded/book

mendelt:
@almindor Sorry, I didn't notice your fix. I accidentally overwrote it.

almindor:
> @almindor Sorry, I didn't notice your fix. I accidentally overwrote it.

NP I only reviewed, no changes.

mendelt:
The link has been updated (twice). Does this require more work from me?

therealprof:
> The link has been updated (twice). Does this require more work from me?

No, sorry, someone should have told our friendly robot about the approval. 😅

PR #449, opened 2020-04-28
[RFC] Minimum Supported Rust Version Revisit

jamesmunns:
[Rendered RFC](https://github.com/rust-embedded/wg/blob/rfc-msrv-2020/rfcs/0000-msrv-2020.md).

Marked as a draft, as not all fields of the RFC have been completed, but the major points are all there.

### DIFF
--- a/rfcs/0000-msrv-2020.md
+++ b/rfcs/0000-msrv-2020.md
@@ -24,7 +24,7 @@ In particular, it has been necessary to balance the (somewhat opposed) concerns 
 
 The following proposed policy aims to be a "good in most cases" solution, to support the largest subset of users and their typical usage practices.
 
-[foundational crates to 1.0](https://github.com/rust-embedded/wg/issues/383)
+[foundational crates to 1.0]: https://github.com/rust-embedded/wg/issues/383
 [has been contentious]: https://github.com/rust-embedded/wg/issues/427
 
 # Detailed design
--- a/rfcs/0000-msrv-2020.md
+++ b/rfcs/0000-msrv-2020.md
@@ -38,10 +38,12 @@ This should get into specifics and corner-cases, and include examples of how the
 
 First, let's start with the "what" of the policy, what are our rules?
 
+> NOTE: These rules apply only to released versions of the WG crates, not necessarily the tip of the `main` branch.
+
 * **Required Versions**: Crates maintained by the Embedded WG must always compile on the three most recent stable semver-minor compiler versions.
     * e.g. if the current version is `1.50.1`, all WG crates must build with `1.48.*`, `1.49.*`, and `1.50.*`.
 * **Minimum Supported Rust Version (MSRV)**: All crates maintained by the Embedded WG must state their **MSRV** in their project README. This must be a version less than or equal to the smallest current **Required Version**.
-    * The MSRV is inclusive of any possible feature of the crate, as well as the MSRV of all dependencies
+    * The MSRV is inclusive of any possible feature of the crate, as well as the MSRV of all dependencies, unless otherwise specified.
 * **Versioning Bump**: When incrementing the **MSRV** of a crate, the crate must make a semver-minor version increment.
     * ex 1: For a crate that is version `1.2.3`, it must then go to version `1.3.0`.
     * ex 2: For a crate that is version `0.4.5`, it must then go to version `0.4.6`.
@@ -51,9 +53,11 @@ First, let's start with the "what" of the policy, what are our rules?
 Then, let's look at the "how" of the policy:
 
 * The **MSRV** should be stated in the project README.
-* The **MSRV** should be tested using CI.
+* The **MSRV** should be tested using CI. Including:
+    * Per-commit/per-PR testing
+    * Periodic CI testing of released crates
 * If a WG crate begins failing due to a dependency using new features of Rust, the WG crate should issue a semver-trivial release setting a maximum version of that dependency. Additionally, the WG crate may consider making a **Versioning Bump** without setting the maximum version of that dependency.
-* Pull requests that *lower* the MSRV will be accepted. However, if the **MSRV** is reduced, a new **Versioning Bump** will be required to reduce it again. Example:
+* Pull requests that *lower* the MSRV will be accepted. However, if the **MSRV** is reduced, a new **Versioning Bump** will be required to increase it again. Example:
     * A WG crate has an **MSRV** of `1.50.0`, and a version of `1.2.5`.
     * A change is made that requires a new version of Rust.
     * The WG crate updates its **MSRV** to `1.52.0`, and its version to `1.3.0`.
@@ -86,14 +90,16 @@ We will need to take the following actions:
 # Drawbacks
 [drawbacks]: #drawbacks
 
-Why should we *not* do this?
+The MSRV process is hugely time consuming to maintain, discuss, and decide. Likely, it currently provides little-to-no-value for most users.
+
+Although the process detailed above is comprehensive as far as we know today, it is likely to have unexpected edge cases that will require refinement.
 
 # Alternatives
 [alternatives]: #alternatives
 
-What other designs have been considered? What is the impact of not doing this?
+The primary alternative to this is to abolish the management of MSRV for Embedded Rust WG projects. This would simplify the effort of managing these releases and projects, however for esoteric users that are restricted to an old version of the compiler, this would require extra effort on their part to maintain or use the crates in question.
 
 # Unresolved questions
 [unresolved]: #unresolved-questions
 
-What parts of the design are still TBD?
+At the moment, there are no open unresolved questions.
--- a/rfcs/0000-msrv-2020.md
+++ b/rfcs/0000-msrv-2020.md
@@ -1,0 +1,99 @@
+- Feature Name: msrv-2020
+- Start Date: 2020-04-28
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+This RFC proposes an update to the [current MSRV policy].
+
+[current MSRV policy]: https://github.com/rust-embedded/wg/blob/8eb6488fdb16e92e70b074acc2fcf249b3edc70b/ops/msrv.md
+
+# Motivation
+[motivation]: #motivation
+
+<!-- Why are we doing this? What use cases does it support? What is the expected outcome? -->
+
+As part of the push to take [foundational crates to 1.0] releases, it has become necessary to be more exact on our guarantees to support versions of the Rust compiler. This discussion [has been contentious] in the past, sparking significant discussions in many meetings.
+
+In particular, it has been necessary to balance the (somewhat opposed) concerns of:
+
+1. Some users may find themselves stuck on old versions of the compiler, due to company restrictions, slow moving distribution/package managers, or regulatory concerns.
+2. We maintain these crates on a volunteer basis, and our time is limited to focus on maintenance
+
+The following proposed policy aims to be a "good in most cases" solution, to support the largest subset of users and their typical usage practices.
+
+[foundational crates to 1.0](https://github.com/rust-embedded/wg/issues/383)
+[has been contentious]: https://github.com/rust-embedded/wg/issues/427
+
+# Detailed design
+[design]: #detailed-design
+
+<!--
+This is the bulk of the RFC. Explain the design in enough detail for somebody familiar
+with the language to understand, and for somebody familiar with the compiler to implement.
+This should get into specifics and corner-cases, and include examples of how the feature is used.
+-->
+
+First, let's start with the "what" of the policy, what are our rules?
+
+* **Required Versions**: Crates maintained by the Embedded WG must always compile on the three most recent stable semver-minor compiler versions.
+    * e.g. if the current version is `1.50.1`, all WG crates must build with `1.48.*`, `1.49.*`, and `1.50.*`.
+* **Minimum Supported Rust Version (MSRV)**: All crates maintained by the Embedded WG must state their **MSRV** in their project README. This must be a version less than or equal to the smallest current **Required Version**.
+    * The MSRV is inclusive of any possible feature of the crate, as well as the MSRV of all dependencies
+* **Versioning Bump**: When incrementing the **MSRV** of a crate, the crate must make a semver-minor version increment.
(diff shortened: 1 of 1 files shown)


rfcs/0000-msrv-2020.md
```diff
@@ -0,0 +1,99 @@
...
+    * ex 2: For a crate that is version `0.4.5`, it must then go to version `0.4.6`.
+* **Best Practices**: When possible, care should be taken to avoid a **Versioning Bump**. However, a **Versioning Bump** is always possible if the above conditions are met, particularly when maintainability of the crate or code simplicity would be improved.
+
+
+Then, let's look at the "how" of the policy:
+
+* The **MSRV** should be stated in the project README.
+* The **MSRV** should be tested using CI.
+* If a WG crate begins failing due to a dependency using new features of Rust, the WG crate should issue a semver-trivial release setting a maximum version of that dependency. Additionally, the WG crate may consider making a **Versioning Bump** without setting the maximum version of that dependency.
+* Pull requests that *lower* the MSRV will be accepted. However, if the **MSRV** is reduced, a new **Versioning Bump** will be required to reduce it again. Example:
```
jonas-schievink, line 56:
```suggestion
* Pull requests that *lower* the MSRV will be accepted. However, if the **MSRV** is reduced, a new **Versioning Bump** will be required to bump it again. Example:
```

rfcs/0000-msrv-2020.md
```diff
@@ -0,0 +1,99 @@
...
+    * ex 1: For a crate that is version `1.2.3`, it must then go to version `1.3.0`.
+    * ex 2: For a crate that is version `0.4.5`, it must then go to version `0.4.6`.
+* **Best Practices**: When possible, care should be taken to avoid a **Versioning Bump**. However, a **Versioning Bump** is always possible if the above conditions are met, particularly when maintainability of the crate or code simplicity would be improved.
+
+
+Then, let's look at the "how" of the policy:
+
+* The **MSRV** should be stated in the project README.
+* The **MSRV** should be tested using CI.
+* If a WG crate begins failing due to a dependency using new features of Rust, the WG crate should issue a semver-trivial release setting a maximum version of that dependency. Additionally, the WG crate may consider making a **Versioning Bump** without setting the maximum version of that dependency.
```
jonas-schievink, line 55: > the WG crate should issue a semver-trivial release setting a maximum version of that dependency

Are there existing projects that do this?

I've argued against this in the last meeting, and I still feel that this is a bad idea for a general policy. To summarize my thoughts:

* Defaulting to this will make any consumers of the library miss patch releases with potentially critical bug fixes
* I believe Cargo's default behavior is to only pull in one semver-compatible version of a crate, but this might break when a maximum version constraint is used (since those are very uncommon). Either way, it will lead to either of these two outcomes:
  * The new release of the dependency will not be used *at all* because of this constraint, even if other dependencies are fine with the new version.
  * Two versions of the dependency will be used by every crate that depends on the WG crate and an unrelated crate that also pulls in the dependency. (This means there's also no way to opt out here; you'll still get the old version even if you use the new version too, which implies that you're fine with a newer Rust version)

As such, I feel like this solution is not fit to be part of the general MSRV policy. It can still make sense to do this, but that's a case-by-case decision. I think the general policy should be to just bump the WG crate's version accordingly with the next release. Also note that we shouldn't feel like we're *forced* to do the new release immediately, as an MSRV bump is just a documentation change, so it can be queued up with other changes.

adamgreig:
We might want to consider regular automated CI runs so we are alerted to dependencies breaking our MSRV without someone having to tell us. At the moment I don't think there's any good notification system, but perhaps something could drop a message into the Matrix or display results on a dashboard or something.

rfcs/0000-msrv-2020.md
jamesmunns: > Are there existing projects that do this?

I'm not aware of any other projects that do this. I don't have a huge amount of visibility, though.

> Defaulting to this will make any consumers of the library miss patch releases with potentially critical bug fixes

That being said, I did try to mitigate this approach by suggesting this in terms of documentation:

> * For users that are sensitive to MSRV changes, we should suggest that they specify the Major and Minor semver version, e.g. version = "1.5.*".
> * For all other users, we should suggest they specify only the Major semver version, e.g. version = "1.*".

This would be the difference between "taking changes, including ones that break MSRV", and "do not take changes that break MSRV".

Without setting an upper bound on the dependency, then we have implicitly increased the MSRV, which at least as described in this RFC, would require a version bump.

> but this might break when a maximum version constraint is used

I would suggest that this might be an issue that should be fixed upstream in Cargo itself, if it is something that "should" work.

> Two versions of the dependency will be used by every crate that depends on the WG crate and an unrelated crate that also pulls in the dependency. (This means there's also no way to opt out here; you'll still get the old version even if you use the new version too, which implies that you're fine with a newer Rust version)

Couldn't this be fixed by an implicit `cargo update -p msrv_breaking_crate` step to force re-consolidation of versions? Not sure if I'm completely understanding.

jamesmunns:
> We might want to consider regular automated CI runs so we are alerted to dependencies breaking our MSRV without someone having to tell us. At the moment I don't think there's any good notification system, but perhaps something could drop a message into the Matrix or display results on a dashboard or something.

Yes! I definitely meant to include nightly CI checks in the RFC. I will add this.

rfcs/0000-msrv-2020.md
jonas-schievink: > For users that are sensitive to MSRV changes, we should suggest that they specify the Major and Minor semver version, e.g. version = "1.5.*".

Does this work in practice or does Cargo have any pitfalls with `*` requirements? Also, the default recommended format should probably just be the usual `version = "1.x.y"`, not `version = "1.*"`.

> Couldn't this be fixed by an implicit `cargo update -p msrv_breaking_crate` step to force re-consolidation of versions? Not sure if I'm completely understanding.

If the WG crate specifies an upper bound, `cargo update` must respect that. If any other crate does not specify that bound and pulls in a second, newer version, `cargo update` will not do anything since it doesn't downgrade unless you tell it to choose a *specific* version of the dependency.

dbrgn:
This looks like a well-balanced RFC to me, I fully support this :slightly_smiling_face: Thanks for the write-up.

rfcs/0000-msrv-2020.md
```diff
@@ -0,0 +1,99 @@
...
+<!--
+This is the bulk of the RFC. Explain the design in enough detail for somebody familiar
+with the language to understand, and for somebody familiar with the compiler to implement.
+This should get into specifics and corner-cases, and include examples of how the feature is used.
+-->
+
+First, let's start with the "what" of the policy, what are our rules?
+
+* **Required Versions**: Crates maintained by the Embedded WG must always compile on the three most recent stable semver-minor compiler versions.
+    * e.g. if the current version is `1.50.1`, all WG crates must build with `1.48.*`, `1.49.*`, and `1.50.*`.
```
dbrgn, line 42: Maybe it should be noted that this applies to releases? I don't see a problem with pre-releases (master or alpha/beta/rc versions) that require newer Rust versions or even nightly.

jamesmunns: Yeah, that was the intent. I'll clarify.

rfcs/0000-msrv-2020.md
```diff
@@ -0,0 +1,99 @@
...
+* **Minimum Supported Rust Version (MSRV)**: All crates maintained by the Embedded WG must state their **MSRV** in their project README. This must be a version less than or equal to the smallest current **Required Version**.
+    * The MSRV is inclusive of any possible feature of the crate, as well as the MSRV of all dependencies
```
adamgreig, line 44: Does this refer to Cargo features, and if so can this be opt-out per feature? For instance, our inline-asm feature doesn't build on any stable Rust, and I could imagine in the future it being useful to feature gate some new functionality that might otherwise cause issues. Maybe we should say the MSRV covers all default features?

rfcs/0000-msrv-2020.md
```diff
@@ -0,0 +1,99 @@
...
+* **Versioning Bump**: When incrementing the **MSRV** of a crate, the crate must make a semver-minor version increment.
+    * ex 1: For a crate that is version `1.2.3`, it must then go to version `1.3.0`.
+    * ex 2: For a crate that is version `0.4.5`, it must then go to version `0.4.6`.
```
Disasm, line 47: I think this should be `0.4.5` -> `0.5.0`. Bumping to `0.4.6` is a patch version bump. Correct me if I'm wrong.

dbrgn: That's a good question. Below 1.0.0, semver says that "anything goes and breakage is always possible". Some people therefore consider 0.X.0 a major release and 0.0.X a minor release (no patch releases, basically everything shifted to the right).

On the other hand even if a crate hasn't reached 1.0.0 yet, we might still want to give some API stability guarantees (a majority of crates is currently still on version 0.x).

adamgreig: [Cargo specifically differs from semver on this point](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements):

> This compatibility convention is different from SemVer in the way it treats versions before 1.0.0. While SemVer says there is no compatibility before 1.0.0, Cargo considers 0.x.y to be compatible with 0.x.z, where y ≥ z and x > 0.

Disasm: Exactly, cargo behavior is the problem here. If MSRV bump is a braking change, then it should change the version in a way cargo considers non-compatible.

Disasm: Just another note: it's enough to change 0.0.4 to 0.0.5 because they are not compatible.

jamesmunns: > Exactly, cargo behavior is the problem here. If MSRV bump is a braking change, then it should change the version in a way cargo considers non-compatible.

I actually did this on purpose. By revving a minor version, this is not objectively a breaking change (as per @jonas-schievink's advice), but DOES at least give folks a way to opt-out of MSRV changes (by specifying `major.minor.*`).

rfcs/0000-msrv-2020.md
Disasm: >>    but this might break when a maximum version constraint is used
>
> I would suggest that this might be an issue that should be fixed upstream in Cargo itself, if it is something that "should" work.

I'm not sure it's a good idea. Pulling a single version of dependency instead of multiple of them is better: multiple versions of the same crate can lead to multiple incompatible interfaces.

adamgreig:
We recently hit a problem with the current MSRV policy in cortex-m, where a dependency (of a dependency of a dependency) had started to use MaybeUninit in a minor release, which had the effect of increasing the MSRV for the already published version of cortex-m. While that problem in general is hard to avoid, at least this new policy would explicitly permit our own updates with increased MSRV to be non-breaking semver changes.

I've tried to summarise the outstanding comments/suggestions/objections to the current RFC text below:

* [ ] "If a WG crate begins failing due to a dependency using new features of Rust, the WG crate should issue a semver-trivial release setting a maximum version of that dependency."
    * https://github.com/rust-embedded/wg/pull/449#discussion_r416821216
    * Causes our dependants to miss updates to that library
    * Might be a breaking change in its own right, since any consumers of our library which were using new features of that dependency would start breaking after updating our library (or they would end up with two copies of the dependency)
    * Perhaps the solution is to just increase our stated MSRV in the next release of our crate, semver-trivial or otherwise, or force a new semver-trivial release with that stated MSRV.
    * Perhaps the wording of our MSRV statement needs to take this into account.
* [ ] Add a mention of periodic automatic CI runs https://github.com/rust-embedded/wg/pull/449#issuecomment-620776257
* [ ] "Crates maintained by the Embedded WG must always compile..."
    * Clarify wording to refer to _releases_ of our crates https://github.com/rust-embedded/wg/pull/449#discussion_r417144653
* [ ] "The MSRV is inclusive of any possible feature of the crate, as well as the MSRV of all dependencies"
    * We probably want to allow some features to require nightly (e.g. inline-asm), which isn't compatible with an MSRV, and the same is true of our dependencies. Perhaps the MSRV could only apply to the default features, or "unless otherwise indicated".

I think the first point is the most contentious. Does anyone else have suggstions? @jamesmunns, would you be able to update the RFC to take some of the points into account?

eldruin:
About point 1, if we relax the policy to be able to raise the MSRV in this situation, I think it would be good to recommend _swiftly_ releasing a new semver-incompatible version, thus making the previous release as somewhat deprecated.
What I would specially like to avoid is that the MSRV needs to be updated several times within a semver-compatible version because the MSRV then really loses its value.

jamesmunns:
@adamgreig I have addresses all comments except for 1. in [this comment](https://github.com/rust-embedded/wg/pull/449#issuecomment-717509516). I feel that this request is incompatible with the policy I describe in this RFC. Specifically:

Regarding @adamgreig's comments:

> Causes our dependants to miss updates to that library

I mean, that's the tradeoff with supporting MSRV in my opinion.

> Might be a breaking change in its own right, since any consumers of our library which were using new features of that dependency would start breaking after updating our library (or they would end up with two copies of the dependency)

That's a fair comment, but I'm not sure what else we can do, outside of specifying exact/maximum versions of every dependency, and updating on every release

> Perhaps the solution is to just increase our stated MSRV in the next release of our crate, semver-trivial or otherwise, or force a new semver-trivial release with that stated MSRV.

The current RFC proposes semver-trivial updates to pin the max acceptable version, and recommends a semver-minor update if we want to continue with a higher MSRV.

> Perhaps the wording of our MSRV statement needs to take this into account.

I'm open to specific suggestions.

Regarding @Disasm's most recent comment above:

> I'm not sure it's a good idea. Pulling a single version of dependency instead of multiple of them is better: multiple versions of the same crate can lead to multiple incompatible interfaces.

I think this is the tradeoff. Update your MSRV to consolidate versions, or don't, and deal with multiple versions. I don't think there is a tradeoff-less option here.

jamesmunns:
Repeating what I said on chat:

I'd like to put a soft limit on proposing any new changes to the MSRV management I described, unless there is something internally inconsistent with the policy I proposed. I understand if people don't *like* the behavior I proposed, but I don't see any other way of "connecting the dots" without just making different tradeoffs.

If someone else wants to edit or change this proposal and submit a competing RFC, or reject this RFC, I support that as an alternative.

This is ready for approval/rejection.

jamesmunns:
Closing in favor of #523. Thanks all!

PR #454, opened 2020-05-12
Rejig repo transfer instructions

therealprof:


adamgreig:
Perhaps this should also mention adding the MSRV to the README?

therealprof:
Can we do this piecemeal wise please? ;)

PR #455, opened 2020-05-14
Hibernating myself

thejpster:


thejpster:
CC @rust-embedded/hal @rust-embedded/resources @rust-embedded/cortex-m

Issue #462, opened 2020-06-28
Update "Organization -> Other projects" section of README.md to list well supported devices


bbrown1867:
As a new Rust embedded user, I would like to see what devices are "well supported" by the Rust community. By "well supported" I mean devices that have a sub-community around them and a more mature HAL, not a single repo created by an individual. For example, the nRF and STM32 microcontrollers appear to fit this definition:
* https://github.com/nrf-rs
* https://github.com/stm32-rs

My understanding is that these communities are not part of the working group, and the working group focuses on supporting architectures rather than devices (e.g. cortex-m, cortex-r, risc-v, etc.). Please correct me if I'm wrong.

Just linking to these communities from the working group README (_Other projects_ seems like a good section) and stating that they are separate communities would help I think. Happy to make the PR.

Disasm:
I'd like to note that STM32 support for different STM32 families varies from mature to "a single repo created by an individual".

bbrown1867:
Perhaps that wasn't the best way to describe it. I just meant that the repos under `stm32-rs` and `nrf-rs` appear to have some sort of community wrapped around them, even if they are developed by one person.

hannobraun:
There's lpc-rs too: https://github.com/lpc-rs

eldruin:
I think more visibility for these high-profile projects would be a good thing.

bbrown1867:
Can I open a PR for this issue? Or does it need to be approved by someone first?

therealprof:
@bbrown1867 Well, we first need to see the proposed changes before approving them. 😅 

Please feel free to PR your ideas and I'll be happy to look into moving your PR forward.

bbrown1867:
Okay great! :) Thought maybe there some issue triage process or something

therealprof:
Well, there is a process for more complex proposals which require some exploration and discussion to decide on technical details. But for more lightweight changes like I expect this one to be I'd start with a proposal and then we can look at whether this is simply a matter of someone rubberstamping it (or suggesting a better alternative) or it warrants a few minutes of discussion in the next meeting or it needs a fully fledged RFC and vote.

This kind of sounds like it go in the first category but we'll see when there's something to see. ;)

bbrown1867:
Closed by #466

PR #466, opened 2020-07-02
Add well supported devices to readme

bbrown1867:
This PR implements the documentation update discussed in #462

### DIFF
--- a/README.md
+++ b/README.md
@@ -482,6 +482,17 @@ These are other projects you may be interested in but that (currently) are not o
 
 ## On going community efforts
 
+### Device specific communities
+
+Several device specific communities exist that are not part of the working group.
+These communities maintain crates for peripheral access, hardware abstraction,
+examples, and more that are specific to a particular family of devices.
+Some of the higher profile ones are listed below.
+
+- [nrf-rs](https://github.com/nrf-rs): Nordic Semiconductor
+- [stm32-rs](https://github.com/stm32-rs): STMicroelectronics
+- [lpc-rs](https://github.com/lpc-rs): NXP Semiconductors
+
 ### [`embedded-hal`]
 
 [`embedded-hal`]: https://github.com/japaric/embedded-hal
--- a/README.md
+++ b/README.md
@@ -487,7 +487,7 @@ These are other projects you may be interested in but that (currently) are not o
 Several device specific communities exist that are not part of the working group.
 These communities maintain crates for peripheral access, hardware abstraction,
 examples, and more that are specific to a particular family of devices.
-Some of the higher profile ones are listed below.
+The list below is not exhaustive and should be updated as device support increases.
 
 - [nrf-rs](https://github.com/nrf-rs): Nordic nRF series of microcontrollers
 - [stm32-rs](https://github.com/stm32-rs): STM 32-bit microcontrollers
--- a/README.md
+++ b/README.md
@@ -487,10 +487,10 @@ These are other projects you may be interested in but that (currently) are not o
 Several device specific communities exist that are not part of the working group.
 These communities maintain crates for peripheral access, hardware abstraction,
 examples, and more that are specific to a particular family of devices.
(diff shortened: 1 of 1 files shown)


therealprof:
Sounds good to me, and I just noticed we should rewrite that whole section anyway since `embedded-hal` is an official part of the WG and not a community effort...

@rust-embedded/all Opinions?

README.md
```diff
@@ -482,6 +482,17 @@ These are other projects you may be interested in but that (currently) are not o
 
 ## On going community efforts
 
+### Device specific communities
+
+Several device specific communities exist that are not part of the working group.
+These communities maintain crates for peripheral access, hardware abstraction,
+examples, and more that are specific to a particular family of devices.
+Some of the higher profile ones are listed below.
```
therealprof, line 490:
```suggestion
```

I don't think we need to claim these are higher profile, there're also plenty other ones around. Maybe encourage people to add them here instead?

jonas-schievink:
We could also make the "awesome embedded Rust" project more visible in the readme and add a section with "bigger" vendor support projects there (eg. anything that is its own GitHub org)

Disasm:
Irrelevant to this PR, I think it's a good idea to have a "hal x peripherals" table with a support level for different peripherals somewhere so that people can match their expectations with the reality before they dig too deep. It will be quite difficult to maintain, though.

README.md
bbrown1867: Agreed, fixed.

bbrown1867:
@therealprof I am happy to remove the `embedded-hal` blurb in this section (2 birds, 1 stone). I noticed it was stale also. What about  `The weekly driver initiative` is that still active?

therealprof:
> We could also make the "awesome embedded Rust" project more visible in the readme and add a section with "bigger" vendor support projects there (eg. anything that is its own GitHub org)

I agree in general, though we're already kind of violating the spirit of "awesome lists" with our interpretation of the rules: https://github.com/sindresorhus/awesome/blob/main/awesome.md

I don't mind mentioning the community driven orgs here as an entry point for interested people so they end up quicker at the right place without shying them away with the information overload on aer. ;)

eldruin:
Yes. I agree the embedded-hal section should definitely not be listed as a community effort.
Some people post every now and again to the weekly driver initiative. Although I agree it is kind of stale, there are a lot of people subscribed so driver announcements can reach a lot of interested people at once. The awesome embedded rust repo is much more active, though.

therealprof:
> I am happy to remove the `embedded-hal` blurb in this section (2 birds, 1 stone). I noticed it was stale also. 

Please let's keep those topics separate. You're very welcome to propose another PR with that change, though. That one is non-controversial. ;)

> What about `The weekly driver initiative` is that still active?

Could use some love and drive but in theory yes.

README.md
```diff
@@ -482,6 +482,17 @@ These are other projects you may be interested in but that (currently) are not o
 
 ## On going community efforts
 
+### Device specific communities
+
+Several device specific communities exist that are not part of the working group.
+These communities maintain crates for peripheral access, hardware abstraction,
+examples, and more that are specific to a particular family of devices.
+The list below is not exhaustive and should be updated as device support increases.
```
adamgreig, line 490:
```suggestion
The list below is not exhaustive and will be updated as device support increases.
```

README.md
```diff
@@ -482,6 +482,17 @@ These are other projects you may be interested in but that (currently) are not o
...
+
+- [nrf-rs](https://github.com/nrf-rs): Nordic nRF series of microcontrollers
+- [stm32-rs](https://github.com/stm32-rs): STM 32-bit microcontrollers
```
adamgreig, line 493:
```suggestion
- [stm32-rs](https://github.com/stm32-rs): STM32 microcontrollers
```

bbrown1867: Fixed, thanks.

README.md
bbrown1867: Fixed, thanks.

PR #474, opened 2020-07-08
No need to spam the old guard

therealprof:
Spamming all current WG members with approval requests should suffice 😅


therealprof:
bors r=eldruin

PR #475, opened 2020-07-09
Mention where to PR Highfive changes

therealprof:


therealprof:
Indeed, however **that** text goes into https://github.com/rust-embedded/wg/blob/master/ops/post-transfer.md ;)

therealprof:
bors r=eldruin

PR #526, opened 2020-11-29
Update MSRV policy

adamgreig:
This PR updates our MSRV policy in line with RFC #523.

### DIFF
--- a/ops/msrv.md
+++ b/ops/msrv.md
@@ -2,48 +2,22 @@
 
 This text documents the MSRV policy used in the crates maintained by the WG.
 
-- The MSRV shall always be strictly smaller than the latest available minor
-  stable release. For example, if the latest stable Rust version was 1.32.1 a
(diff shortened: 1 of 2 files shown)


thalesfragoso (approved):
Great, thanks!

bors r+

therealprof:
@adamgreig Do you have a good recipe to add scheduled builds to CI? Would be great to add that to our documented procedures as well.

eldruin:
@therealprof you can add an schedule entry which uses cron syntax. For example:
```yaml
on: 
  push:
  pull_request:
  schedule:
    - cron: 0 1 * * 1
```
(Taken from [here](https://github.com/eldruin/driver-examples/blob/1419cd2c32a204393a4040cad157e56c2f6cc692/.github/workflows/build.yml#L5))

therealprof:
@eldruin Thanks, I was aware that can be done that way. I was hoping that someone had a refined version already worked out which we could massage into out docs (and all WG repos). 😅

PR #577, opened 2021-09-28
Mention support of RP2040 microcontrollers

jannic:
While it's still a work in progress, the community at [rp-rs](https://github.com/rp-rs) has built usable Rust crates for the RP2040 microcontroller, which should be mentioned in the list of device specific communities.

### DIFF
--- a/README.md
+++ b/README.md
@@ -481,6 +481,7 @@ The list below is not exhaustive and will be updated as device support increases
 - [lpc-rs](https://github.com/lpc-rs): NXP LPC microcontrollers
 - [imxrt-rs](https://github.com/imxrt-rs): NXP i.MX RT microcontrollers
 - [esp-rs](https://github.com/esp-rs): Espressif Systems microcontrollers
+- [rp-rs](https://github.com/rp-rs): Raspberry Silicon RP2040 microcontroller 
 
 ### [`embedded-hal`]
 
--- a/README.md
+++ b/README.md
@@ -481,7 +481,7 @@ The list below is not exhaustive and will be updated as device support increases
 - [lpc-rs](https://github.com/lpc-rs): NXP LPC microcontrollers
 - [imxrt-rs](https://github.com/imxrt-rs): NXP i.MX RT microcontrollers
 - [esp-rs](https://github.com/esp-rs): Espressif Systems microcontrollers
-- [rp-rs](https://github.com/rp-rs): Raspberry Silicon RP2040 microcontroller 
+- [rp-rs](https://github.com/rp-rs): Raspberry Pi microcontrollers including the RP2040
 
 ### [`embedded-hal`]
 


README.md
```diff
@@ -481,6 +481,7 @@ The list below is not exhaustive and will be updated as device support increases
 - [lpc-rs](https://github.com/lpc-rs): NXP LPC microcontrollers
 - [imxrt-rs](https://github.com/imxrt-rs): NXP i.MX RT microcontrollers
 - [esp-rs](https://github.com/esp-rs): Espressif Systems microcontrollers
+- [rp-rs](https://github.com/rp-rs): Raspberry Silicon RP2040 microcontroller
```
adamgreig, line 484: Makes sense to me! Do you think "Raspberry Pi microcontrollers" might fit better?

```suggestion
- [rp-rs](https://github.com/rp-rs): Raspberry Pi microcontrollers
```

jannic: I copied the term from the linked page, which was just cleaned up by @thejpster.
There was a discussion in the pull request at https://github.com/rp-rs/rp-hal/pull/137#discussion_r717232134 with the conclusion that the term "Raspberry Pi" doesn't apply to the microcontroller itself.

adamgreig: I see! Still, the company is "Raspberry Pi" and it's their microcontrollers, right?

image

jannic: Yes, right, and I wouldn't mind adding "Pi", I just left it out for consistency with the linked page.
However, I would keep the RP2040 for now, as I think this is the most common term the microcontroller is called. And even if some day the family will grow, I guess the name RP2040 will stick for a while.

adamgreig: Maybe "Raspberry Pi microcontrollers including the RP2040"?

jannic: The trademark rules could be another reason to leave out the "Pi":
https://www.raspberrypi.org/trademark-rules/
"We own the following trademarks: The words “Raspberry Pi” [...] Any use of the Raspberry Pi Marks must be accompanied by the acknowledgement “Raspberry Pi is a trademark of the Raspberry Pi Foundation”."

jannic: But then, the suggested usage would not be the name of some third party product, but we would actually refer to the company Raspberry Pi and their microcontroller. I guess this wouldn't be a trademark violation. (With my very limited knowlege of law...)

adamgreig: Yes, I'm sure it would be fine for us to say "Raspberry Pi microcontrollers including RP2040" here.

adamgreig (approved):
Thanks, LGTM!

bors merge

README.md
thejpster: Yeah, as discussed in our repo, they are quite clear to refer to the PCBs as Raspberry Pi's but not the chips. Some Raspberry Pis have Broadcom silicon and some have Raspberry Silicon. Our support is for the chip, not the PCB, so I think it should be as written.

https://www.raspberrypi.org/blog/the-journey-to-raspberry-silicon/

Issue #647, opened 2022-11-22
Tracking Issue: Rust Embedded on Mastodon


jamesmunns:
Following up from chat, we probably want to have some kind of presence on Mastodon/fediverse.

The two main open questions:

**Who is going to run it?**

IMO (@jamesmunns): whoever runs it should be on the Resources team. We should also figure out how to share access to the account to avoid it getting abandoned, etc.

* @mygnu has volunteered

We may want an existing RE member, maybe @thejpster or @adamgreig, to help mentor whoever runs the account.

@jamesmunns can help with "open source brand" style and coming up with general practices to be consistent with the twitter and cohost accounts.

**Which server should it be on?**

There are open discussions on this. Chat discussed:

* hachyderm.io
* fosstodon.org

mygnu:
I'm happy to help also in favour of hachyderm.io (biased cos I already have an account there) I believe @thejpster is also there. We should definitely have some guidelines around how to operate such account.

thejpster:
I'm @thejpster@hachyderm.io.

I understant Kris at Hachyderm.io, like many volunteer instance mods, has a *lot* going on right now. So whoever we approach we should do it with care and understanding.

thejpster:
Hachyderm has had a big infra upgrade. Could be a good home now the dust has settled.

adamgreig:
Sounds good to me! I've set up @rustembedded@hachyderm.io, and am just sorting out having it go to the resources team email address. As far as I could tell there's no reason not to just create the account their; registrations are open. I don't think there's any way to share ownership beyond just sharing the password, which I'm happy to do.

We should work out some basic principles for operating the accounts as @jamesmunns suggested; perhaps something like:

1. Post about specific announcements from the working group, e.g. to tell people about some major new release, a breaking change, etc
2. Post WG blog posts
3. Repost/share people offering and seeking jobs, with some rate limit that doesn't feel too spammy and will have to adjust based on how many requests we get
4. Repost/share people's posts tagged/mentioning us about embedded rust, same deal with the rate limit
5. Repost/share other major Rust related posts, e.g. from the official Rust account

How does that sound?

mygnu:
Sounds good to me

thejpster:
yup, good starter for ten.

jamesmunns:
Opened https://github.com/rust-embedded/wg/pull/651, feel free to add any feedback or suggestions there.

PR #651, opened 2022-12-07
Add initial draft of social media ops

jamesmunns:
This is a first draft of a general "social media ops" outline.

CC #647.

CC @adamgreig, @thejpster, @mygnu, I could use a hand with "how to mastodon"!

### DIFF
--- a/ops/social-media.md
+++ b/ops/social-media.md
@@ -1,0 +1,66 @@
+# Social Media Guide
+
+The [Resources Team] currently maintains three official social media accounts:
+
+* [Twitter: @rustembedded](https://twitter.com/rustembedded)
+    * Maintained by [@jamesmunns]
+* [Cohost: @rustembedded](https://cohost.org/rustembedded)
+    * Maintained by [@adamgreig] and [@jamesmunns]
+* [Mastodon: @rustembedded@hachyderm.io]
+    * Maintained by [@adamgreig]
+
+[Resources Team]: https://github.com/rust-embedded/wg#the-resources-team
+[@jamesmunns]: https://github.com/jamesmunns
+[@adamgreig]: https://github.com/adamgreig
+[Mastodon: @rustembedded@hachyderm.io]: https://hachyderm.io/@rustembedded
+
+## Operational Guide
+
+The purpose of the social media accounts is to spread awareness of topics related to embedded rust.
+
+Social Media accounts will typically:
+
+1. Post about specific announcements from the working group, e.g. to tell people about some major new release, a breaking change, etc
+2. Post WG blog posts
+3. Repost/share people offering and seeking jobs, with reasonable rate-limiting (e.g. one post/week per person/group), see below for more info
+4. Repost/share people's posts tagged/mentioning embedded rust, with reasonable rate-limiting (e.g. one post/week per person/group)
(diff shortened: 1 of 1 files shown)


ops/social-media.md
```diff
@@ -0,0 +1,66 @@
...
+
+### For Cohost
+
+On Cohost, we ask that people include the [#rustembedded](https://cohost.org/rc/tagged/rustembedded) tag on their posts.
+
+The cohost account follows this tag, and pulls posts from the stream.
+
+### For Mastodon
+
+TODO: James doesn't know how this works
```
mygnu, line 53: I'd say Mastodon is much more 'live' in a way that its purely time based, if you share or boost (retoot) something it is broadcasted at that time. if anyone missed it or scroll passed it it won't show up in their feed unless they explicitly look back on it. No of people liking a post has no effect on the timeline.

For Tagging you can create a column (web app) that will show toots with your selected tags, i.e. #rustembedded, that can be kept an eye on separately and boosted as required. 
asking folks to mention @rustembedded would also be good.

jonathanpallant: @rustembedded@hachyderm.io, specifically.

jamesmunns: (either of you) please feel free to suggest language for this section - whatever makes sense for Mastodon!

adamgreig:
LGTM, thanks for putting it together!

ops/social-media.md
mygnu: Here is one attempt
```
On Mastodon, we ask that people "@ mention" the rustembedded account, e.g. include "@rustembedded@hachyderm.io" in a toot.

On Mastodon we can follow #rustembedded tag directly in a column that will update with all tagged local (to hachyderm.io) toots and the instances it federates with.
```

mygnu:
Just wondering if we are waiting on something for this?

ops/social-media.md
esden, line 53: On mastodon hashtags are a very important tool. You should also specify what tag to use for Rust Embedded related posts. You can also specify that you will boost the posts with a specific tag for example `#embeddedrustboost`. You can describe what the rules are either in your profile or as a pinned post. (on mastodon you can have more than one pinned post)

esden:
I would suggest that a bsky account is also added to the official Rust Embedded account list. If that is done I would suggest creating "starter packs" with accounts that are recommended to follow for additional Rust Embedded related content. Additionally I would also create dedicated feed of Rust Embedded content. Keep also in mind that the AT protocol allows for domain based nicknames. So if there is a Rust Embedded specific domain I would highly recommend linking that to the bsky account.

ops/social-media.md
```diff
@@ -0,0 +1,66 @@
+# Social Media Guide
+
+The [Resources Team] currently maintains three official social media accounts:
+
+* [Twitter: @rustembedded](https://twitter.com/rustembedded)
+    * Maintained by [@jamesmunns]
+* [Cohost: @rustembedded](https://cohost.org/rustembedded)
+    * Maintained by [@adamgreig] and [@jamesmunns]
```
jannic: This is outdated. Cohost was shut down in 2024, and twitter is now X and I seriously doubt we should be present there.

PR #653, opened 2022-12-21
Rename to embedded-alloc

Jzow:
CC [#56](https://github.com/rust-embedded/embedded-alloc/pull/56).

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,7 +173,7 @@ The Cortex-M team develops and maintains the core of the Cortex-M crate ecosyste
 
 Projects maintained by this team.
 
-- [`alloc-cortex-m`]
+- [`embedded-alloc`]
 - [`cortex-m-quickstart`]
 - [`cortex-m-rt`]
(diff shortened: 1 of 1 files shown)


reitermarkus:
As I already asked in https://github.com/rust-embedded/embedded-alloc/pull/56#discussion_r1041727500: Should this be maintained by the HAL team now that it is not specific to Cortex-M anymore?

Jzow:
> Should this be maintained by the HAL team now that it is not specific to Cortex-M anymore?

@adamgreig  @eldruin @dirbaio , Is it necessary to re enable the RFC process?

eldruin:
From an outsider's perspective, I agree it may make sense to move it to the HAL team.
Let's discuss this in tomorrow's meeting.

eldruin (approved):
We agreed to keep it maintained by the cortex-m at least for the time-being in the last two weekly meetings so let's merge this as-is.
Thank you!
bors r+

PR #686, opened 2023-07-12
Update cortex-a team crates: drop cortex-a and register-rs, add aarch64-cpu

adamgreig:
cortex-a and register-rs are both deprecated; cortex-a was replaced by aarch64-cpu while register-rs is replaced by tock-registers (not part of the wg).


eldruin (approved):
Sounds good to me, thanks!

PR #727, opened 2024-01-11
Leave tools team, join HAL, libs and embedded Linux teams.

reitermarkus:
I initially joined the tools team to maintain `cross`, which isn't in this org anymore and I am not really using the other tools. I mostly use `heapless`, `embedded-hal`, `critical-section` and the various embedded Linux crates these days, and it makes sense to maintain the crates I actually use.

### DIFF
--- a/README.md
+++ b/README.md
@@ -332,7 +332,6 @@ embedded targets themselves.
 - [@adamgreig]
 - [@burrbull]
 - [@Emilgardis]
-- [@reitermarkus]
 - [@ryankurte]
 - [@therealprof]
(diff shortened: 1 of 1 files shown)


therealprof:
Would you mind splitting this up into multiple PRs to simplify the voting which has to happen per team?

adamgreig (approved):
Sorry this got left out, I've updated to just leaving tools team since there's separate PRs for the others.

adamgreig:
Thank you for all your help with cross!

PR #751, opened 2024-04-24
assets: Add Sticker Design File

jamesmunns:
This is the sticker design I ordered on Diginate. The red outline was the cut line.


eldruin (approved):
Thanks James!

PR #824, opened 2025-03-05
Add JP to Arm team.

jonathanpallant:
I'd like to join the Arm team, to continue the work on the cortex-ar and cortex-r-rt crates.

Work will slow down now my demo is ready for EW2025, but I can work on them in the background. I usually won't be attending the weekly meetings due to time constraints.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,6 +133,7 @@ and microcontrollers, including Cortex-A, Cortex-R, and Cortex-M.
 
 - [@adamgreig]
 - [@berkus]
+- [@jonathanpallant]
 - [@nchong-at-aws]
 - [@newAM]
 - [@raw-bin]


adamgreig:
Thank you! As usual this will require a vote from the arm team members:

- [x] [@adamgreig]
- [ ] [@berkus]
- [ ] [@nchong-at-aws]
- [ ] [@newAM]
- [ ] [@raw-bin]
- [ ] [@thalesfragoso]
- [ ] [@therealprof]

jonathanpallant:
https://github.com/rust-embedded/wg/blob/master/rfcs/0136-teams.md says

> Voting majority is defined as having more approvals than abstentions and zero unresolved concerns.

However it doesn't specify a time-frame.

At what point is this considered accepted?

berkus:
And I'm pretty sure this hasn't passed because folks simply did not see this issue.

We need a voting method that tracks presence too.

thalesfragoso (approved):
Sorry for the delay, glad to have you on board.

